PR #13, opened 2024-03-07
Fix transform_str_to_dict and make it more flexible.

### LINKED ISSUE #4: Extend chem_formula utility functions
@t-reents:

_I realized that the current implementation of the utility functions in the `cehem_formula` module doesn't support certain aspects. For example, `transform_str_to_dict` doesn't support formulas containing parentheses. The problem is that we don't raise any error in case parentheses are present, and the parsed result is incorrect. Therefore, we should decide which cases will be supported and add some validation for the remaining ones._

_Another group of formulas that cause problems are those that aren't grouped, e.g. "HOH" wouldn't be correctly parsed to `{"H": 2, "O": 1}`. As this kind of formula is also supported in AiiDA's `StructureData().get_formula()`, we should think about supporting it as well (which would be quite easy in this case)._

_However, we should really think about what to support. Pymatgen implements a lot of logic to capture all of these special cases in `pymatgen.core.Composition`, so I'd opt for not completely reinventing the wheel._

hdsassnick:
`transform_str_to_dict` supports now:

* recurring element symbols
* Different kinds of nested brackets.

Closes #4 

@t-reents any other case that needs to be supported?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+Version 0.2.0
+=============
+
+**Fixes:**
+
+* ``utils.chem_formula.transform_str_to_dict`` supports now recurring elements and nested brackets (`PR #13 <https://github.com/aim2dat/aim2dat/pull/13>`_).
+
+
 Version 0.1.0
 =============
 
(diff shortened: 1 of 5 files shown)


t-reents (approved):
Thanks @hdsassnick! 
I think this covers everything that I had in mind. 

I don't have further comments, looks good.

hdsassnick:
> Thanks @hdsassnick! I think this covers everything that I had in mind.
> 
> I don't have further comments, looks good.

Thanks, I will fix the last test and then merge it

PR #23, opened 2024-04-03
Add option in subplot to adjust y labels

Scemp:
For plotting for example partial charges, the y label usually is not aligned since the y range varies for each species. Matplotlib has a function for this, and I added the property `subplot_align_ylabels` to the `_Baseplot` class to call the function.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,10 @@ Version 0.2.0
 
 * ``utils.chem_formula.transform_str_to_dict`` supports now recurring elements and nested brackets (`PR #13 <https://github.com/aim2dat/aim2dat/pull/13>`_).
 
+**Enhancement**
+
+* ``plots.base_plot`` added new property `subplot_align_ylabels` to adjust y labels in subplots for backend `matplotlib`.
+
 
 Version 0.1.0
 =============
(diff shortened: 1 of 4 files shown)


t-reents:
Thanks @Scemp! Looks good and can be merged in principle. Just one comment/question: Have you also checked those cases where you would benefit from this automatic alignment in Plotly? I'm aware that we have the same logic for the `tight_layout` which is only supported in `Matplotlib`. Nonetheless, I think it would be nice if both backends support as many of the properties as possible. 
In case something similar exists in `Plotly`, it would be great to add this as well. If it doesn't, and the labels are also not properly aligned in such cases (that were problematic in `Matplotlib`), we could think if there is an easy way to implement such a function ourselves?
If this is not straightforward, we can for sure merge this first and maybe keep track of these properties in another issue or similar.

Scemp:
Hi everyone!
@hdsassnick that is a good point and will be updated soon. I will include it to the next push when we figured out what we need to change to include `Plotly`.
@t-reents I checked the documentation for `Plotly` and could not find anything similar to this addition. Also, I tried some workarounds and unfortunately, I was not able to adjust the distances from the y label to the axes. 
Regarding `subplot_tight_layout`, I was also checking if I can find something similar but without success. To implement something from our side will be not easy. Meanwhile, I checked the properties `subplot_hspace` and `subplot_wspace`. This only works between 0 and 1 for `Plotly`. For `Matplotlib` you can adjust the numbers according to your wanted spacings. Maybe we should check each property for `Plotly`.
All the best!

hdsassnick:
Great thank you, @scemp and @t-reents . As the unsuccessful checks are unrelated to this PR, I am merging this and added #24

PR #26, opened 2024-04-25
Added new file formats to write structure to file

Scemp:
For post-processing with Zeo++ we need certain file formats. So, I included `.cssr`, `.v1` and `.cuc` to the `to_file` function.


aim2dat/strct/strct.py
```diff
@@ -621,10 +621,43 @@ def from_aiida_structuredata(
...
+            f.close()
+        elif file_path.endswith(".cuc"):
+            f = open(file_path, "w")
+            f.write(f"Processing: {self.label}\n")
+            f.write("Unit_cell: ")
+            f.write(" ".join(map(str, self.cell_lengths)) + " ")
+            f.write(" ".join(map(str, self.cell_angles)) + "\n")
+            for el, pos in zip(self.elements, self.scaled_positions):
+                f.write(f"{el} {pos[0]} {pos[1]} {pos[2]}\n")
+            f.close()
```
t-reents, line 657: I would put this into a separate (private) function as well or even better into a new module in `ext_interfaces`. I would say it's up to you whether you put those if-elif statements into one function `_write_structure_to_file` or into separate private functions for each file format which are then called by `_write_structure_to_file`. The latter might be useful in case this could be extended to other file formats in the future (maybe this is not relevant for Zeo++, I'm not familiar with it). 

By doing so, we can simply use `ase` as the default backend or call the `zeo++` backend if one of the file formats is specified. This would make things a bit cleaner and more compact.

aim2dat/strct/strct.py
```diff
@@ -621,10 +621,43 @@ def from_aiida_structuredata(
...
+            f.write(" SPGR = " + self.determine_space_group()["space_group"]["int_symbol"] + "\n")
+            f.write(f"{len(self.positions)} 0\n")
+            f.write(f"0 {self.label}\n")
+            for i, el_pos in enumerate(zip(self.elements, self.scaled_positions)):
+                f.write(
+                    f"  {i+1} {el_pos[0]} {el_pos[1][0]} {el_pos[1][1]} {el_pos[1][2]} "
+                    + " ".join(map(str, 9 * [0]))
+                    + "\n"
+                )
+            f.close()
```
t-reents, line 639: I only mention it here but this would be valid for all the different file formats. Instead of explicitly opening and closing the file for each file format, you could append all the different lines for the input file to a list, e.g. `output`. Afterwards, you could use something like

```
with open(file_path, 'w') as file:
    file.write('\n'.join(output))
```
to finally write the file.

t-reents (commented):
Thanks @Scemp, I left some general comments.

aim2dat/strct/strct.py
Scemp: Thanks for the suggestion. I will fix it.

aim2dat/strct/strct.py
Scemp: Hi Timo,
You are totally right by adding a private function. So, the function is only called if the formats are needed. 

But with the new package `ext_interfaces` I am not sure since we could have two options, and I am not sure which one I prefer at the moment:

1.  We could add a new module `zeo++` and extend the module over the time. Maybe we want to import the outputs from the program later for further investigation. Right now, I am not sure if this is necessary.
2. By having a private function not connected to a certain module, we can always extend the function in case someone requires a certain file structure.

t-reents: Yes, I see your point. First of all, I just realized that `ext_interfaces` would be the wrong place as we provide interfaces to other python libraries there. Instead, we should use `io`. 

Without thinking too much about it, I could think of it in the following way: If you think that parsing of the outputs etc. might be useful at some point, I would create a module `zeo` in `io` and implement a function that takes care of transforming a structure into the appropriate file format. 
If this is not the case, one could think of adding a module (not 100% where to add it, probably also `io`) where we put a function for each file format separately. This might be useful in case we want to support certain file formats that don't belong to a certain library/code (not sure how often this would be relevant, having in mind that `ase` supports a bunch of different formats).

In either of these options, we could have a mapping between file format and corresponding backend function. If the file format doesn't belong to one of the "custom" ones, we simply use ase.

If we choose option 1, this could look like the following:
```
format_mapping = {
    'cssr': 'zeo++',
    'v1': 'zeo++',
    'cuc': 'zeo++'
}

backend_module = _return_ext_interface_modules(format_mapping.get(file_format, 'ase_atoms'))
backend_module._write_structure_to_file(structure, file_path)
```

In case we go with the separate function for each format, one wouldn't need the backend_module and could directly provide the function to convert the structure in the mapping.

Scemp: Hi Timo,

thanks for the suggestion. I added a new module `zeo` in `io` since we might add some functions later to import the output of `Zeo++` later. Also, `ext_interfaces` seems to be only for external python libraries and `Zeo++` is a program not based on python. Therefore, I assume the latest version is the best option for now.

t-reents:
Thanks @Scemp, looks good now. As discussed, we should only add a small test before merging

t-reents (approved):

hdsassnick:
Great, nice work. Thanks @Scemp and @t-reents . I am merging it :)

PR #28, opened 2024-05-07
Fix bugs in io.fhi_aims.read_band_structure.

### LINKED ISSUE #27: 2 bugs in fhi-aims `read_band_structure` function
Pull request #20 introduced two bugs into the `read_band_structure` function:

1. If a non-soc calculation is performed a ValueError is always triggered. It should be only triggered if `soc=True`.
2. If other output files are in the same folder the routine will read all files with suffix `.out` due to a too unspecific regex pattern.

hdsassnick:
Fixed the bugs described in #27 and extended the tests such that the issues are detected.

Closes #27


t-reents (approved):
Looks good to me.

PR #29, opened 2024-05-07
List of `Structure` objects as input to `StructureOperations`

### LINKED ISSUE #17: Allow list of ``Structure`` objects as input for the constructor of ``StructureOperations``
It would be nice to have a shortcut for: 
``strct_op = StructureOperations(StructureCollection([structure1, structure2])`` 
such that this works: 
``strct_op = StructureOperations([structure1, structure2])``

t-reents:
Fixes #17 

The `StructureOperations` class takes only a `StructureCollection` as an input at the moment. Here we allow the user to pass a list of `Structure` objects to initialize `StructureOperations`.

@hdsassnick `StructureCollection` also accepts a list of dictionaries. I only added support for a list of `Structure`s to initialize `StructureOperations`, as this is somehow the main object to represent a structure. In case you prefer to also allow a list of dictionaries, I will add this as well.

### DIFF
--- a/aim2dat/strct/structure_collection.py
+++ b/aim2dat/strct/structure_collection.py
@@ -41,7 +41,7 @@ class StructureCollection:
         List of ``Structure`` or dict objects.
     """
 
-    def __init__(self, structures: Union[list, None] = None):
+    def __init__(self, structures: Union[List[Union[Structure, dict]], None] = None):
         """Initialize object."""
         self._structures = []
         if structures is not None:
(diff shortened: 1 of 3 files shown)


hdsassnick:
Thanks @t-reents , looks good to me.

To have things consistent I would just check whether `structures` is a `StructureCollection` or list and pass the list on to a new `StructureCollection` without any further checks. So whenever we choose to have a different input format for `StructureCollection` it would be also usable in `StructureOperations`, what do you think?

t-reents:
Hi @hdsassnick, totally agree. I adjusted it accordingly.

hdsassnick (approved):
Thanks, merging this one too.

PR #31, opened 2024-05-09
Voronoi analysis - fixed order

t-reents:
The order of the neighboring sites obtained from the Voronoi analysis is fixed to a certain sorting in this commit. The background of this change is that the unsorted results caused some problems with the tests.


hdsassnick (approved):
Thanks @t-reents , I will merge since the failing tests are unrelated (I will fix the MP interface as soon as I find some time...).

PR #32, opened 2024-05-23
Fix `smearing`: add sampling width

t-reents:
Currently, the distance between adjacent x-values is not taken into account, which leads
to unreasonable smearing results.

Here, the sampling width is added as an optional argument, ensuring backward compatibility. Nonetheless, one could think of making this argument non-default.

The aforementioned unreasonable results are visualized in the following figure. I use a lorentzian smearing with a FWHM of 5eV. As shown, the effect of smearing in the old result without the proper sampling width is clearly too small. 

image

### DIFF
--- a/aim2dat/fct/smearing.py
+++ b/aim2dat/fct/smearing.py
@@ -13,7 +13,7 @@ def _gaussian(x: np.array, sigma: float) -> np.array:
     x : np.array
         x-values.
     sigma : float
-        Scale parameter.
+        Scale parameter corresponding to the standard deviation.
 
     Returns
     -------
@@ -32,7 +32,7 @@ def _lorentzian(x: np.array, sigma: float) -> np.array:
     x : np.array
         x-values.
     sigma : float
-        Scale parameter.
+        Scale parameter corresponding to the full-width at half-maximum (FWHM).
 
     Returns
     -------
@@ -47,7 +47,11 @@ AVAILABLE_SMEARING_METHODS = {"gaussian": _gaussian, "lorentzian": _lorentzian}
 
 
 def apply_smearing(
-    y: np.array, sigma: float = 0.5, radius: int = None, method: str = "gaussian"
+    y: np.array,
+    sampling_width: float = 1.0,
+    sigma: float = 0.5,
+    radius: int = None,
+    method: str = "gaussian",
 ) -> np.array:
     """Apply smearing to a dataset. Different smearing methods can be specified.
 
@@ -55,8 +59,10 @@ def apply_smearing(
     ----------
     y : np.array
         y-values of dataset.
+    sampling_width : float
+        Sampling width of the x-axis, i.e. distance between adjacent x-values. Defaults to 1.0.
     sigma : float
-        Scale parameter. Defaults to 0.05.
+        Scale parameter. Defaults to 0.5.
     radius : int
         Radius of the kernel.
     method : str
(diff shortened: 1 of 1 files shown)


t-reents:
Hi @hdsassnick, good point. I think that it would make sense to add it there as well. One could think of setting it to `None` per default and determining it from the provided data, in case no explicit value is specified by the user. As shown in the example, I think that it is especially relevant for spectroscopic/DOS data. 

If we don't want to determine it automatically, we can instead think of setting it to a reasonable default value.

hdsassnick:
> Hi @hdsassnick, good point. I think that it would make sense to add it there as well. One could think of setting it to `None` per default and determining it from the provided data, in case no explicit value is specified by the user. As shown in the example, I think that it is especially relevant for spectroscopic/DOS data.
> 
> If we don't want to determine it automatically, we can instead think of setting it to a reasonable default value.

Ok, then I would suggest two options:

* We merge this as it is and add an issue to address the plot classes in a later PR,
* or please feel free to update the `_SmearingMixin` class. I am actually just seeing that we already have a `smearing_delta` parameter there, I guess we just have to check how this relates to the sampling width, right? I would then prefer to have a uniform interface and depreciate `smearing_delta`...

t-reents:
I quickly checked the impact of `smearing_delta` and I think this parameter and `sampling_width` are not mutually exclusive. The `smearing_delta` takes care of binning and summing the datapoints which can also change the scale of the data. However, it can also be set to `None` and the original data without binning is used. 
This being said, might it actually make sense to keep both options?

If yes, one could think of setting `sampling_width` to `smearing_delta` in case it's not `None` and otherwise set `sampling_width` to a default or automatically determined value.

hdsassnick:
> I quickly checked the impact of `smearing_delta` and I think this parameter and `sampling_width` are not mutually exclusive. The `smearing_delta` takes care of binning and summing the datapoints which can also change the scale of the data. However, it can also be set to `None` and the original data without binning is used. This being said, might it actually make sense to keep both options?
> 
> If yes, one could think of setting `sampling_width` to `smearing_delta` in case it's not `None` and otherwise set `sampling_width` to a default or automatically determined value.

I also think that one of the two parameters should be sufficient. So, I would introduce `sampling_width` and depreciate `smearing_delta` (add a warning for now).

t-reents:
In principle I agree, but how would you proceed with the binning and summation?

On the one hand, I think it makes sense that the values are aggregated like this, in case the user provides a `sampling_width` that is larger than the actual difference between adjacent x-values.
On the other hand, this kind of aggregation and summation changes the scale of the y-values, and especially the area. In case the user is actually interested in these properties, I'm not sure whether this should be the default behaviour.

Therefore, I was thinking in my previous comment to keep both attributes, to keep the binning logic and the `sampling_width` of the data separately.

The main point is: I think it makes sense that the `sampling_width` is set to `smearing_delta` (in case it is not `None`), but I'm not sure whether the other way around is 100% reasonable as well, i.e. always setting `smearing_delta` and applying the binning, in case one specifies `sampling_width` . If we did the binning and calculate the mean per bin instead of the sum, I'd say yes. But in the current version, I'm not 100% sure. 
I've just made a test and both approaches are actually equivalent if we replace the summation by the mean value, in that case, we could easily rely on the `sampling_width` only.

Long story short, I'd suggest to adapt the aggregation logic to take the mean instead of the sum and implement the `sampling_with` as the only parameter.

hdsassnick:
sorry, I guess I would need to run a few tests myself to fully understand the impact of both parameters.

But I think your solution sounds reasonable, we should just make sure that it is backwards compatible, maybe we can  keep the original behavior with `smearing_delta` and depreciate it. What do you think?

PR #33, opened 2024-05-24
Fix FHI-aims pdos parser for higher quantum numbers and added test.

hdsassnick:
In some cases FHI-aims would include g quantum numbers in the pDOS output files which were not considered in the original implementation of the parser.

This PR fixes this issue in the parser and the DOSPlot class.


t-reents (approved):
Looks good to me.

PR #34, opened 2024-05-28
Unify structure manipulation interface and introduce Structure.to_dict function 

hdsassnick:
This pull request includes four changes:

* Unifies the interface of the internal and external structure manipulation methods.
* Allows a structure manipulation function to return a Structure object instead of a dictionary.
* Implements a central `to_dict` function to easily transform a `Structure` object into a python dictionary (replacing the "custom" transformation in each manipulation function).
* Adds an example of an external manipulation function in the documentation.


aim2dat/strct/ext_manipulation/add_functional_group.py
```diff
@@ -115,20 +114,15 @@ def add_functional_group(
...
-            "pbc": structure["pbc"],
-            "is_cartesian": True,
-            "wrap": wrap,
-            "positions": list(structure["positions"]),
-            "cell": structure["cell"],
-            "elements": list(structure["elements"]),
-        }
+        new_structure = structure.to_dict()
+        new_structure["elements"] = list(new_structure["elements"])
+        new_structure["positions"] = list(new_structure["positions"])
```
t-reents, line 119: `structure.to_dict()` also contains the `kinds` attribute. Therefore, we are causing an inconsistency between the `elements` and `kinds` in this case. This problem wasn't present before, as the dictionary was initialized from scratch.

aim2dat/strct/strct_manipulation.py
```diff
@@ -75,14 +76,8 @@ def substitute_elements(
...
-        new_structure = {
-            "label": structure.label,
-            "elements": list(structure["elements"]),
-            "positions": structure["scaled_positions"],
-            "is_cartesian": False,
-            "pbc": structure["pbc"],
-            "attributes": {},
-        }
+        new_structure = structure.to_dict(cartesian=False)
+        new_structure["elements"] = list(new_structure["elements"])
```
t-reents, line 80: Same as above, we are changing the `elements` in the following loop but don't modify the `kinds`. This might not be the expected behavior.

t-reents (changes requested):
Looks good to me in general. I just added two small comments concerning the `elements` and `kinds`.

aim2dat/strct/ext_manipulation/add_functional_group.py
hdsassnick: Thanks and I agree, I have implemented a fix and adapted the test case.

aim2dat/strct/strct_manipulation.py
hdsassnick: In principle yes, but `kinds` are not necessarily related to `elements`. I am a bit hesitant to remove additional information from the `Structure` object that may be needed (e.g. if a specific site is marked regardless of the element it hosts, etc.).

Maybe an additional parameter should be implemented to cover both use cases, something like `remove_kind` or `erase_kind`? That way, at least from the doc-strings the behavior should be clear?

t-reents: Yes, totally agree. I was thinking about something similar, to allow the user to choose. 

Having in mind the use-case that you described, where the kinds are used to label certain sites, it definitely makes sense to keep them.
However, when using the kinds to distinguish magnetic moments, it might be unexpected to find `Co1` and `Co2` kinds in a `Ni` structure, just as an artificial example.

Scemp (approved):
Looks also good to me! Thanks @hdsassnick and @t-reents

PR #35, opened 2024-05-28
Added 'atomic_radius' method to calculate coordination numbers.

hdsassnick:
This PR adds:
* The atomic radii determined in https://doi.org/10.1039/C9RA07327B to the `utils.element_properties.get_atomic_radius` function.
* And implements the calculation of the coordination environment based on the sum of the corresponding atomic radii.


aim2dat/strct/strct_coordination.py
```diff
@@ -254,6 +267,43 @@ def _coord_calculate_n_nearest_neighbours(
...
+    dist_matrix,
+    radius_type,
+    atomic_radius_delta,
+):
+    el_idx_sc = indices_sc.index(site_idx)
+    # min_dist = np.amin(np.delete(dist_matrix[site_idx], el_idx_sc))
+    position = np.array(structure.positions[site_idx])
+    site_radius = get_atomic_radius(structure.elements[site_idx], radius_type=radius_type)
+    if site_radius is None:
+        site_radius = get_atomic_radius(structure.elements[site_idx], radius_type="covalent")
```
t-reents, line 287: I think that this logic is not completely correct, or better to say, doesn't work as intended. You only check whether the radius is `None`, but the `ase.data` references return `nan` in case the value is not available. This can be seen in the following example for `Sc`.

```
from aim2dat.utils.element_properties import get_atomic_radius

radius = get_atomic_radius('Sc', 'vdw')
print(radius, radius is None) 
# nan False
```

As expected, the calculation of the coordination number for this scandium structure using the `vdw` radii would also fail:

```
from ase.build import bulk
from aim2dat.strct import Structure

sc = Structure.from_ase_atoms(bulk('Sc', 'hcp', a=3.3, c=5.2))
sc.calculate_coordination(method='atomic_radius', radius_type='vdw')

{'distance_avg': {},
 'distance_stdev': {},
 'distance_max': {},
 'distance_min': {},
 'nrs_avg': {('Sc', 'Sc'): 0},
 'nrs_stdev': {('Sc', 'Sc'): 0.0},
 'nrs_max': {('Sc', 'Sc'): 0},
 'nrs_min': {('Sc', 'Sc'): 0},
 'sites': [{'Sc': 0,
   'element': 'Sc',
   'kind': None,
   'position': [0.0, 0.0, 0.0],
   'neighbours': [],
   'total_cn': 0,
   'min_dist': 0.0,
   'max_dist': 0.0,
   'avg_dist': 0.0},
  {'Sc': 0,
   'element': 'Sc',
   'kind': None,
   'position': [-1.2952601953960163e-17, 1.9052558883257646, 2.6],
   'neighbours': [],
   'total_cn': 0,
   'min_dist': 0.0,
   'max_dist': 0.0,
   'avg_dist': 0.0}]}
```

aim2dat/utils/element_properties.py
```diff
@@ -54,26 +54,39 @@ def _check_element(element):
...
     if radius_type == "covalent":
         radius = covalent_radii[el_number]
     elif radius_type == "vdw":
         radius = vdw_radii[el_number]
+    elif radius_type == "chen_manz":
+        file_path = os.path.dirname(__file__) + "/data_files/atomic_radii.yaml"
+        radius = load_yaml_file(file_path)[radius_type][element]
     else:
         raise ValueError(f"Radius type '{radius_type}' not supported.")
     return radius
```
t-reents, line 92: Maybe you could adapt the logic here, something alike:

```suggestion
    if radius is None or np.isnan(radius):
        return None
    return radius
```

`numpy` would need to be imported in this case.

t-reents (changes requested):
Thanks @hdsassnick. I think there is a small bug, see comments above.

hdsassnick:
Hi @t-reents , sorry I guess my PRs yesterday were a bit rushed. Thanks for the catch. I implemented your suggestion and added a test case for the `get_atomic_radius` function.

t-reents (approved):
Thanks for the changes, good to go now!

Scemp (approved):
Thanks, @hdsassnick, for implementing the approach. It looks good to me!

Issue #6, opened 2024-03-06
Support for new CP2K versions


hdsassnick:
So far only up cp2k v9.1 is supported by the work chains. It would be great to support also the newer versions of the code.

PR #36, opened 2024-05-29
Revise CP2K parsers

hdsassnick:
Sorry, this PR got quite large...
It removes the legacy parser (`aim2dat.io.cp2k.legacy_parser`) class and replaces it with a more flexible regex-based function and classes to process regex patterns.

The test cases for the stdout file were reduced to a reasonable number which should have a positive impact on test runtimes.

This is also a preparation for #6


aim2dat/utils/dict_tools.py
```diff
@@ -0,0 +1,116 @@
...
+        elif isinstance(helper_dict, dict):
+            helper_dict[parameter] = {}
+            helper_dict = helper_dict[parameter]
+        else:
+            can_add = False
+
+    if can_add:
+        helper_dict[parameter_tree[-1]] = value
+    else:
+        raise ValueError("Cannot add value to dictionary.")
```
t-reents, line 39: I don't have the time for a full review now but I started to play around with this function. As discussed, I think I found some ways to break the current implementation.

1. The following example shows that there is a problem in case the second to last `key` is not a dictionary. In the last iteration of the loop, `helper_dict` is set to `1` which causes the final assignment to fail. The reason is that there was no additional iteration where `can_add` would be set to `False`.

Solution: One needs to check after the final iteration whether `helper_dict` is actually a `dict`
```
test_dict = {'1': {'1.1': 1}}
dict_set_parameter(test_dict, ['1', '1.1', '1.2'], 1)

-------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[56], line 2
      1 test_dict = {'1': {'1.1': 1}}
----> 2 dict_set_parameter(test_dict, ['1', '1.1', '1.2'], 1)
      3 print('old', test_dict)

File ~/project/dev_aim2dat/git/aim2dat/aim2dat/utils/dict_tools.py:37, in dict_set_parameter(dictionary, parameter_tree, value)
     34         can_add = False
     36 if can_add:
---> 37     helper_dict[parameter_tree[-1]] = value
     38 else:
     39     raise ValueError("Cannot add value to dictionary.")

TypeError: 'int' object does not support item assignment
```

2. In this case, I added additional keys so that there is an iteration where `can_add` is actually set to `False`. Here, the error is raised because we try to check whether a certain key is present in an `int`. 

Solution: Before checking that the `key` is present, one needs to check that `helper_dict` is still a `dict`
```
test_dict = {'1': {'1.1': 1}}
dict_set_parameter(test_dict, ['1', '1.1', '1.2', '1.3'], 1)

-------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[68], line 2
      1 test_dict = {'1': {'1.1': 1}}
----> 2 dict_set_parameter(test_dict, ['1', '1.1', '1.2', '1.3'], 1)
      3 print('old', test_dict)

File ~/project/dev_aim2dat/git/aim2dat/aim2dat/utils/dict_tools.py:28, in dict_set_parameter(dictionary, parameter_tree, value)
     25 can_add = True
     27 for parameter in parameter_tree[:-1]:
---> 28     if parameter in helper_dict:
     29         helper_dict = helper_dict[parameter]
     30     elif isinstance(helper_dict, dict):

TypeError: argument of type 'int' is not iterable
```

To sum up, both examples actually suggest the same solution. I drafted a possible solution that fixes those issues and that is also more compact due to fewer if-elif statements. Moreover, we can already raise the error once we encounter a state where `helper_dict` is not a `dict`, so there is no need to continue the loop.

I tested several examples and both versions agree (in those cases where the previous version worked successfully).

```suggestion
    helper_dict = dictionary

    for parameter in parameter_tree[:-1]:
        if isinstance(helper_dict, dict):
            helper_dict = helper_dict.setdefault(parameter, {})
        if not isinstance(helper_dict, dict):
            raise ValueError("Cannot add value to dictionary.")

    helper_dict[parameter_tree[-1]] = value
```

One final aspect: I realized that this function would silently overwrite a value. Just to double-check if this is the intended behaviour.

hdsassnick: Well, actually the function was and still is intended to be used internally by routines of the library (originally for the aiida work chains). But your improvements make sense, thanks :)

And overwriting is intended.

t-reents:
I quickly tested my suggestions, but it would be good if you also check that I didn't break anything

aim2dat/utils/dict_tools.py
```diff
@@ -0,0 +1,109 @@
...
+            if parameter in helper_dict:
+                helper_dict = helper_dict[parameter]
+            else:
+                helper_dict[parameter] = {}
+                helper_dict = helper_dict[parameter]
+        else:
+            raise ValueError("Cannot create nested dictionary.")
+
+
+def dict_merge(a, b, path=None):
```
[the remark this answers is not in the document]
hdsassnick: Yes, at least so far, there was no use case where this was a problem..

PR #37, opened 2024-06-02
Feature/add molecule fct

hdsassnick:
This PR includes:

- The new function `add_structure` replacing the `add_functional_group` function (which is depreciated from now on) and a new test case.
- Bugfix in `external_manipulation_method` decorator.

`add_structure` allows a `Structure` object as input, H2O is added as the first predefined structure representing a molecule. `host_indices` allows a list of indices defining adsorption sites situated between several sites. `dist_constraints` offers a way to control the orientation of the added molecule or functional group.

The implementation is probably not yet ideal but improves on the original function..


hdsassnick:
@t-reents and @Scemp, just introduced a small name change to implement some other functions to "merge" `Structure` objects later on. 

The failed tests are unrelated, will fix those later if the fails persist...

aim2dat/strct/ext_manipulation/add_functional_group.py
```diff
@@ -41,11 +25,15 @@ def add_functional_group(
     """
     Add a functional group or an atom to a host site.
 
+    Notes
+    -----
+        This function is depreciated and will be removed soon. Please use
+        ``aim2dat.strct.ext_manipulation.add_structur`` instead.
```
t-reents, line 31:
```suggestion
        ``aim2dat.strct.ext_manipulation.add_structure`` instead.
```

t-reents (changes requested):
Sorry for being late 😢  Just found a very minor typo. 
Apart from that, good to go.

aim2dat/strct/ext_manipulation/add_structure.py
```diff
@@ -0,0 +1,294 @@
...
+    Returns
+    -------
+    aim2dat.strct.Structure
+        Structure with attached functional group.
+    """
+    guest_strct, guest_strct_label = _check_guest_structure(guest_structure)
+    if isinstance(host_indices, int):
+        host_indices = [host_indices]
+
+    if all(idx < len(structure) for idx in host_indices) and guest_index < len(
```
Scemp, line 110: I was wondering where the else statement is? This should give an error since the index is out of range, but this error will appear anyway if you try to run the function. Therefore, I would leave the if statement out

hdsassnick: This is expected behavior. The function should return None if the indices are out of range.

t-reents: Since this if statement is quite long, it might make sense to reverse the logic. Just handle the out of range case first (check it in an if statement) and return `None`. If this is not triggered, the function will anyway go ahead and return the new structure in the end. In this way, this large block is not indented and one doen't forget the purpose of the if statement

Issue #7, opened 2024-03-06
Add occupancy to Structure class


hdsassnick:
While it is difficult to computationally handle fractional occupations it allows us to implement schemes to "simplify" structures towards full occupations.

hdsassnick:
Closed by #39

Issue #5, opened 2024-03-06
Add support for magnetic moments


hdsassnick:
This issue is to keep track of some tasks which are related to adding support for magnetic moments in the `Structure` class.

- [ ] Add a `magnetic_moments` property in `Structure`
- [ ] Adapt the external interfaces like ASE, pymatgen, spglib and possibly also AiiDA
- [ ] Adjust the affected methods like `determine_spacegroup` (add optional kwarg and call the appropriate method to handle magnetic symmetries) and comparison methods.
- [ ] Set spglib >= 2.0 as dependency (they changed the interface for magnetism)

PR #39, opened 2024-06-20
Feature/add site attributes

### LINKED ISSUE #38: How to best deal with site properties.
There are several attributes that can be assigned to the atoms of a molecule or crystal, e.g.

- Partial occupations (see #7)
- Magnetic moments (see #5)
- Partial charges
- Oxidation states
- Atomic masses
- ...

I see three options:

- Add a dictionary to each site (stored as a list with same dimension as elements) which can have arbitrary contents.
- Add specific properties to `Structure` likewise to `kinds`
- A combination of the first two options. The question then would be, which ones should have fixed names and which to leave out as generic "site properties".

@t-reents and @Scemp , any thoughts or ideas on this?

hdsassnick:
Closes #38

It is also gives the framework for #5 and #7


t-reents:
Just gave it a quick look, no specific comments from my side.
I also saw your issue #38 but I don't have a strong opinion on that atm. The current implementation seems reasonable and we can any we keep it as the internal logic and might explicitly expose certain properties at some point, e.g. via additional properties that simply access the internal `dict`

PR #45, opened 2024-07-02
Change from aiida_script to aim2dat

Scemp:
Not sure whether `aim2dat.strct.strct.Structure` or `aim2dat.strct.Structure`.

### DIFF
--- a/aim2dat/strct/strct.py
+++ b/aim2dat/strct/strct.py
@@ -562,7 +562,7 @@ class Structure(AnalysisMixin, ManipulationMixin):
 
         Returns
         -------
-        aiida_scripst.strct.Structure
+        aim2dat.strct.strct.Structure
             Structure.
         """
         backend_module = _return_ext_interface_modules("ase_atoms")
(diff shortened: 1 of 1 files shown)


aim2dat/strct/strct.py
```diff
@@ -562,7 +562,7 @@ def from_file(cls, file_path: str, attributes: dict = None, label: str = None) -
 
         Returns
         -------
-        aiida_scripst.strct.Structure
+        aim2dat.strct.strct.Structure
```
hdsassnick, line 565: We should best have in the documentation the highest-level import (this way we don't break the interface if we change the actual location of the class at some point...)
```suggestion
        aim2dat.strct.Structure
```

aim2dat/strct/strct.py
```diff
@@ -595,7 +595,7 @@ def from_ase_atoms(
...
```
hdsassnick, line 598:
```suggestion
        aim2dat.strct.Structure
```

aim2dat/strct/strct.py
```diff
@@ -627,7 +627,7 @@ def from_pymatgen_structure(
...
```
hdsassnick, line 630:
```suggestion
        aim2dat.strct.Structure
```

aim2dat/strct/strct.py
```diff
@@ -660,7 +660,7 @@ def from_aiida_structuredata(
...
```
hdsassnick, line 663:
```suggestion
        aim2dat.strct.Structure
```

hdsassnick (changes requested):
Great, thank you for catching this.

PR #53, opened 2024-07-11
Feature/add structure random

hdsassnick:
Adds new function `add_structure_random` that adds the guest structure at random position and orientation.


aim2dat/strct/ext_manipulation/add_structure.py
```diff
@@ -25,6 +26,83 @@
...
+            cell[d][d] = max_pos[d] + threshold
+    else:
+        min_pos = np.zeros(3)
+        cell = np.array(structure.cell)
+
+    random.seed(a=random_state)
+    max_tries = 1000
+    for _ in range(max_tries):
+        guest_positions = np.array(guest_strct["positions"])
+        rot_v = np.array([random.random(), random.random(), random.random()])
```
t-reents, line 84: Just out of curiosity, is there a reason for not using `numpy.random.random_sample(3)`, to directly generate an array of random values (also between 0 and 1)?

t-reents (approved):
Just a small comment, otherwise it's good to go

aim2dat/strct/ext_manipulation/add_structure.py
hdsassnick: I guess, I prefer using the standard python functions when possible. In this case it is also important to have reproducible results for the tests (I guess this would also work with the numpy functions somehow...).

PR #54, opened 2024-07-12
Support new spglib version

hdsassnick:
Revises the spglib interface to conform with the newly implemented Dataset classes in version 2.5.

Keywords the dictionary are changed to conform with the spglib nomenclature.


hdsassnick:
We may also consider a new release soon to prevent errors related to the new spglib version. @t-reents and @Scemp , what do you think?

t-reents:
HI @hdsassnick!

I think we can definitely do a minor release, i.e. 0.2, as we've already added a set of new features. If the changes related to `spglib` are really breaking, we'll have to consider a major release, i.e. 1.0. In that case, we should consider which additional features/open issues we'd like to include.
I'd say that we have to think if there are additional issues/planned adjustments that introduce breaking changes. Those should be included as well so that we don't have to do another major release soonish, and consider that major release as the first stable version.

hdsassnick:
I would not yet consider version 1.0 as this would represent a stable release. I don't consider this package yet mature enough to do that. As far as I understand, we can also release version 0.2.0, even if we introduce breaking changes, based on: https://semver.org/#spec-item-4

This PR is indeed introducing breaking changes as it alters the dictionary keys of the space group/symmetry related output. The rationale behind it is to not have an "additional" set of keys representing the same data. Since I would consider spglib a standard, I thought we should conform to their nomenclature. This way the interface should be more intuitive for users. 

But I am also fine with reverting it and mapping the attributes onto our custom keys (as was done prior to this PR). I am happy to discuss pros and cons. In any case, breaking changes should happen better now than later since the user base is very, very small at the moment...

hdsassnick:
To go forward, I have added a milestone for the next release and added a few issues (mainly minor things) that should ideally be included.

As for the version, I have put v0.2.0 for now, but I am open for suggestions or discussion. @t-reents , @Scemp please feel free to tune in...

Scemp:
Also, I would suggest releasing a new version, because some major changes happened until now. From my point of view, version v0.1.1 would only consider bug fixes and enhancements, and since we added some new features, I would assume v0.2.0 would match the current version. As you also mentioned, v1.0.0 is not done yet. We still have some additional work to do until we can release a stable version. 

I will start working on the milestone soon. Thus, we can release version v0.2.0 soon.

t-reents:
Sorry for being late to the party 😅 @hdsassnick totally agree now where you were mentioning the reference. I only connected breaking changes with major releases, but since we are still in a 0.x, I agree that it's fine to have breaking changes and agree to do a 0.2 release.

I'll also catch up with the reviews next week, got sick the last days

Scemp (approved):
I have not found any discrepancies in the code and would therefore give the green light to push.

t-reents (approved):
Gave it a quick look and I don't have any additional comments, also considering that @Scemp has  already checked it, thanks!

PR #60, opened 2024-07-23
Fix utils.chem_formula.reduce_formula

### LINKED ISSUE #55: `reduce_formula` function cannot handle float numbers appropriatly.
The function to reduce chemical formulas should have as an additional precondition that the resulting reduced formula does not contain float numbers. Otherwise, `compare_formulas` will not work in all cases.

hdsassnick:
Closes #55

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -31,12 +31,13 @@ Version 0.2.0
 
 **Depreciations:**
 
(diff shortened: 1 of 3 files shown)


hdsassnick:
Failed test is unrelated and will be fixed at a later point if it persists.

t-reents (approved):
LGTM!

Scemp (approved):
Looks good to me.

PR #61, opened 2024-07-25
Speed-up identification of duplicate structures.

hdsassnick:
This PR adds a comparison of the chemical formula of the two structures before other more expensive checks. This way, only structures with the same stoichiometry are considered as duplicate structures.

### DIFF
--- a/aim2dat/strct/structure_operations.py
+++ b/aim2dat/strct/structure_operations.py
@@ -48,11 +48,11 @@ def _create_index_combinations(confined, strct_c):
     for idx0 in range(strct_c_len):
         for idx1 in range(min_idx, max_idx):
             if (
-                utils_cf.compare_formulas(
+                idx0 != idx1
+                and (idx1, idx0) not in pairs
+                and utils_cf.compare_formulas(
                     strct_c[idx0].chem_formula, strct_c[idx1].chem_formula, reduce_formulas=True
                 )
-                and idx0 != idx1
-                and (idx1, idx0) not in pairs
             ):
                 pairs.append((idx0, idx1))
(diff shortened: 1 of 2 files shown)


aim2dat/strct/structure_operations.py
```diff
@@ -44,7 +47,13 @@ def _create_index_combinations(confined, strct_c_len):
...
     for idx0 in range(strct_c_len):
         for idx1 in range(min_idx, max_idx):
-            if idx0 != idx1 and (idx1, idx0) not in pairs:
+            if (
+                utils_cf.compare_formulas(
+                    strct_c[idx0].chem_formula, strct_c[idx1].chem_formula, reduce_formulas=True
+                )
+                and idx0 != idx1
+                and (idx1, idx0) not in pairs
+            ):
```
t-reents, line 56:
```suggestion
            if (
                idx0 != idx1
                and (idx1, idx0) not in pairs
                and utils_cf.compare_formulas(
                    strct_c[idx0].chem_formula, strct_c[idx1].chem_formula, reduce_formulas=True
                )
            ):
```

The improvement in terms of speed will be small since the comparison of the formulas is quite cheap. Nonetheless, the comparison will be skipped if the indices are identical or the pair already exists.

t-reents (changes requested):

hdsassnick:
Thanks a lot @t-reents for the swift review of the PR (and the other PRs as well...)
I included your suggestion, should be ready to be merged now.

t-reents (approved):

Scemp (approved):
Loooks good to me.

PR #63, opened 2024-07-25
Unify behavior of structure manipulation methods.

### LINKED ISSUE #52: Structure manipulation methods should carry on site attributes from the input structures
The user might expect that site attributes and kinds are carried on within a structure manipulation. So far, this is not always the case. Ideally, we should have a uniform way to handle this and document it.

hdsassnick:
In general, all methods maintain now ``attributes``, ``site_attributes`` and ``kinds`` throughout structure manipulation.

Closes #52


hdsassnick:
This would be the last PR for the new release. After all I decided to move the revision of the smearing Mixin class to a later release due to time constraints...

@t-reents and @Scemp ,if you agree I would make the new release once all PRs are merged...

aim2dat/strct/ext_manipulation/add_structure.py
```diff
@@ -284,7 +280,12 @@ def add_structure_coord(
 def _check_guest_structure(guest_strct: Union[Structure, str]) -> dict:
     if isinstance(guest_strct, Structure):
         label = "" if guest_strct.label is None else guest_strct.label
-        return {"elements": guest_strct.elements, "positions": guest_strct.positions}, label
+        return {
+            "elements": guest_strct.elements,
+            "positions": guest_strct.positions,
+            "kinds": [None] * len(guest_strct) if guest_strct.kinds is None else guest_strct.kinds,
```
t-reents, line 286: As a general comment, have we already considered to return `[None] * len(elements)` in case `kinds` is not set in a `Structure` object? We always have these if-else statements and I was wondering if there is a reason against this kind of default value for `kinds`? It is also handled in a similar way in `ASE`: https://wiki.fysik.dtu.dk/ase/_modules/ase/atoms.html#Atoms.get_tags. Not to say that we have to do it like that, but just that this might be something to consider

For sure, this would require a revision of all the structural methods and doesn't have to be done right now, I just wanted to put this topic on the table.

t-reents (approved):
LGTM, only added a general comment.

You can continue with the release as suggested!

aim2dat/strct/ext_manipulation/add_structure.py
```diff
@@ -358,6 +355,36 @@ def _add_mol(
...
+    }
+    if len(guest_strct["site_attributes"]) > 0:
+        for site_attr in guest_strct["site_attributes"].keys():
+            if site_attr not in new_structure["site_attributes"]:
+                new_structure["site_attributes"][site_attr] = [None] * len(
+                    new_structure["elements"]
+                )
+    for idx, (el, kind, pos) in enumerate(
+        zip(guest_strct["elements"], guest_strct["kinds"], guest_strct["positions"])
+    ):
```
Scemp, line 377: idx is not used, I would suggest:

    for (el, kind, pos) in (
        zip(guest_strct["elements"], guest_strct["kinds"], guest_strct["positions"])
    ):

Scemp (approved):
I only found a variable which was not in use. But this is only a quick fix. Apart from that, it looks good to go! Thanks, @hdsassnick

aim2dat/strct/ext_manipulation/add_structure.py
hdsassnick: This is a good point. I opened an issue (#66) and will address this still before the next release.

aim2dat/strct/ext_manipulation/add_structure.py
hdsassnick: Thanks for the catch

aim2dat/strct/ext_manipulation/add_structure.py
t-reents: Great!

hdsassnick:
@Scemp if you agree, I would merge this one now and then we include the implementations of your PR, what do you think?

Scemp:
@hdsassnick sounds good. I am happy to merge this and include the other implementations later. So you are good to go!

PR #65, opened 2024-08-02
Added H2 as option for additional molecule

Scemp:
These changes were made:

- Hydrogen `H2.yaml` was added to use hydrogen as an adsorption molecule
- negation was made and an `IndexError` raised if host index and guest index out of range
--> thus the lines do not have to be intended anymore and increases readability 
- added arguments for all implemented `calculate_coordination` methods


hdsassnick:
This PR conflicts with #63. In principle, I wouldn't mind in which PR we implement the changes, however #63 also addresses issues in other functions.

@Scemp , would it be ok if I merge #63 first and then based on those changes we can redesign this PR?

aim2dat/strct/ext_manipulation/add_structure.py
```diff
@@ -108,11 +108,14 @@ def add_structure_coord(
     host_indices: Union[int, List[int]] = 0,
     guest_index: int = 0,
     guest_structure: Union[Structure, str] = "CH3",
+    guest_dir: List[float] = [0.0, 0.0, 0.0],
```
hdsassnick, line 111: We could choose `None` as default?

hdsassnick (commented):
On first sight, I am a bit confused that you needed to change the test files...
Is this due to the `_rescale_bond_length` function?

aim2dat/strct/ext_manipulation/add_structure.py
Scemp: I would leave it as `List[float] = [0.0, 0.0, 0.0]`, so the user sees immediately the input structure.

aim2dat/strct/ext_manipulation/add_structure.py
```diff
@@ -183,55 +189,124 @@ def add_structure_coord(
...
         host_indices = [host_indices]
 
-    if all(idx < len(structure) for idx in host_indices) and guest_index < len(
-        guest_strct["elements"]
-    ):
-        # Shift guest site to [0.0, 0.0, 0.0]
-        guest_strct["positions"] = [
+    if max(host_indices) >= len(structure) or guest_index >= len(guest_strct):
+        idx = max(host_indices)
+        raise IndexError(
```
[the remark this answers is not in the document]
t-reents: Hi @hdsassnick After briefly thinking about it, I can see reasons for and against each approach. What about allowing the user to choose? Similar to methods in `pandas` where you can also specify in certain methods that errors are either `raised` or `ignored`. 

What do you think about a similar approach? In some cases one might want to see the errors (also in `StructureOperations`) whereas they should be ignored in other use-cases

Scemp:
> On first sight, I am a bit confused that you needed to change the test files... Is this due to the `_rescale_bond_length` function?

I did change the tests because the **bond length** from the guest molecule to the host atom and the **orientation** of the guest molecule will be calculated by default now. This changes the outcome of the function. I should not have changed the tests in the first place. Therefore, I optimized reference data now.

aim2dat/strct/ext_manipulation/add_structure.py
t-reents: If I may add my 2 cents, I'm also in favour of `None`. I think the format is quite clear from the docstring and `None` seems to be a more intuitive value since it will be overridden anyway.

aim2dat/strct/ext_manipulation/add_structure.py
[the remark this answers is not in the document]
hdsassnick: Hi @t-reents, thanks for tuning in :) . I completely agree. I would prefer the option where an error is always raised in the manipulation functions but have a property defined in `StructureOperations` to ignore them.

But I think, this should be done in a separate PR and adjust all manipulation functions at once.

aim2dat/strct/ext_manipulation/add_structure.py
hdsassnick: Also, we should avoid mutable types as input parameters in general. So if we keep [0,0,0], a tuple would be safer.

hdsassnick:
> > On first sight, I am a bit confused that you needed to change the test files... Is this due to the `_rescale_bond_length` function?
> 
> I did change the tests because the **bond length** from the guest molecule to the host atom and the **orientation** of the guest molecule will be calculated by default now. This changes the outcome of the function. I should not have changed the tests in the first place. Therefore, I optimized reference data now.

No worries, just to be sure that we don't introduce any bugs/unintentional behavior. I am not sure I fully understand yet, how the behavior of the function changes with this PR. Can we reproduce the original test results? If yes, I would adapt the input parameters but leave the reference data for the tests as they are and add additional test cases for the new features. What do you think, @Scemp?

aim2dat/strct/ext_manipulation/add_structure.py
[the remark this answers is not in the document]
t-reents: Definitely, I'd also do it separately

Scemp:
> > > On first sight, I am a bit confused that you needed to change the test files... Is this due to the `_rescale_bond_length` function?
> > 
> > 
> > I did change the tests because the **bond length** from the guest molecule to the host atom and the **orientation** of the guest molecule will be calculated by default now. This changes the outcome of the function. I should not have changed the tests in the first place. Therefore, I optimized reference data now.
> 
> No worries, just to be sure that we don't introduce any bugs/unintentional behavior. I am not sure I fully understand yet, how the behavior of the function changes with this PR. Can we reproduce the original test results? If yes, I would adapt the input parameters but leave the reference data for the tests as they are and add additional test cases for the new features. What do you think, @Scemp?

I do understand your concern, at the moment it is not implemented to reproduce the original test results if we give more than one host index as an input. My idea is sketched below. The idea is to get the bond length to one of the host atoms. Before, it was the distance between the mean of the host atom positions and the guest molecule. Thus, the length between the host and the guest was always greater than the actual bond length.

Based on the orientation of the molecule, it is possible to reproduce the original test files. Before, each guest molecule was arranged around the x-axis. Thus, if the input `guest_dir = [1.0, 0.0, 0.0]` is given, we can create at least the orientation of the guest molecule.

bond_length_example

aim2dat/strct/ext_manipulation/add_structure.py
Scemp: Would you suggest to change it to the following then:
`guest_dir: Union[None, List[float]] = None`

hdsassnick:
> I do understand your concern, at the moment it is not implemented to reproduce the original test results if we give more than one host index as an input. My idea is sketched below. The idea is to get the bond length to one of the host atoms. Before, it was the distance between the mean of the host atom positions and the guest molecule. Thus, the length between the host and the guest was always greater than the actual bond length.
> 
> Based on the orientation of the molecule, it is possible to reproduce the original test files. Before, each guest molecule was arranged around the x-axis. Thus, if the input `guest_dir = [1.0, 0.0, 0.0]` is given, we can create at least the orientation of the guest molecule.
> 
> bond_length_example

Thanks for the clarification. To move forward, I would suggest to merely adjust the input parameters for the test cases with just one host index. And for the new features, we can add one or two more tests if needed?

Unfortunately, I am quite constrained this week, but next week I can also help with the necessary changes. Also, we should not forget to adjust the interface to the ``add_functional_group`` function to maintain backwards compatibility..

hdsassnick:
Hi @Scemp, I have adapted the function to reproduce the original behavior and I think I also fixed a bug as well.

Please double check if the newly implemented enhancements still work as intended. I think we don't need a new test case (I adapted the existing ones to cover most of the code) but if you feel differently please feel free to add one.

If @t-reents has no further comments, this PR would then finally be ready to be merged :)

aim2dat/strct/ext_manipulation/add_structure.py
```diff
@@ -201,22 +201,24 @@ def add_structure_coord(
...
-            radius_type=radius_type,
-            atomic_radius_delta=atomic_radius_delta,
-            econ_tolerance=econ_tolerance,
-            econ_conv_threshold=econ_conv_threshold,
-            voronoi_weight_type=voronoi_weight_type,
-            voronoi_weight_threshold=voronoi_weight_threshold,
-            okeeffe_weight_threshold=okeeffe_weight_threshold,
-        )
-        guest_dir = -_derive_bond_dir(guest_index, guest_strct_coord)
+    if guest_dir is None:
```
Scemp, line 204: Thanks for pointing this out!

t-reents (approved):
No further comments from my side

PR #71, opened 2024-08-29
Fix slicing in `StructureOperations`

t-reents:
When using a slice to index `StructureOperations`, an exception is raised as we try to call `self.get_structure`. This is only defined for a `StructureCollection`. Changing it to `self.structures.get_structure` fixes the bug.


hdsassnick (approved):
Thanks for the catch, @t-reents.

t-reents:
@hdsassnick The failing tests are only related to the MP interface, so I guess I can merge nonetheless?

hdsassnick:
Yes, let's merge.

Issue #68, opened 2024-08-28
Remove `key` from `ManipulationMixin` and `AnalysisMixin`


t-reents:
In the current version, all methods in `ManipulationMixin` and `AnalysisMixin` have a `key` argument. This is set to `None` in case the method is called on a `Structure` method. We only use it to select the desired structures in case the method is called on `StructureOperations`. 

I've been thinking about the following suggestion from time to time and finally write it down :D 

While the adjustment of the arguments doesn't really cost performance, personally, I'm not 100% happy with it (although I understand and remember our previous considerations). It feels a bit weird to have an additional argument in all the functions that is overridden for all the applications inside `Structure`.

https://github.com/aim2dat/aim2dat/blob/1952d240c726266c916a2f2383288a007ac79ff0/aim2dat/strct/mixin.py#L49-L51

My suggestion would be the following:
What about changing the syntax of `StructureOperations` from 

```
strct_ops = StructureOperations(structures)
strct_ops.any_method([1, 2, 4], *args)
```
to

```
strct_ops = StructureOperations(structures)
strct_ops[[1, 2, 4]].any_method(*args)
```

We already provide indexing and slicing which creates a new `SturctureOperations` object containing the subset of structures. Moreover, it neither complicates the code nor makes it longer. 
Since we don't store anything in the object but only in the `Structure` objects, returning slices is also not a problem. The result is returned and not stored anyway.

In this way, we could remove the key arguments and get an even more uniform interface between `Structure` and `StructureOperations`.

@hdsassnick @Scemp Comments are more than welcome, happy to discuss.

hdsassnick:
Thanks @t-reents for this elegant solution, I am just not sure, how the implementation would look like and whether it can be parallelized?

t-reents:
1. So the general implementation wouldn't change that much (at least in my dreams haha). We basically move the selection of which structures to consider for a certain operation from the function call to the indexing. Therefore, we could remove the following part and simply iterate over all the structures in the `StructureOperations`, as the selection was already done before

t-reents:
Hi @hdsassnick 

[...]

Issue #70, opened 2024-08-29
Calculating `f-fingerprints` in advance seems to be slower


t-reents:
I'm working on implementing a parallelized version of the comparison methods and started with the `f-fingerprint` on a local branch: https://github.com/t-reents/aim2dat/commit/1760619dcecca4203dead17032cf8f5fbb295c5c

The changes generally work. However, I realized that the following part is somehow slowing down the comparison:

https://github.com/aim2dat/aim2dat/blob/1952d240c726266c916a2f2383288a007ac79ff0/aim2dat/strct/structure_operations.py#L433-L435

As it should speed up the execution (all fingerprints are calculated in advance and could be reused, especially when parallelizing), I also implemented it in the new version for the comparison. Thereby, I noticed that it seems to slow things down, also in case of `find_duplicates_via_ffingerprint`. 

You can simply reproduce this behaviour by commenting out the lines shown above and test the `find_duplicates_via_ffingerprint` with and without. Important to note, this can be done on a local copy of `main` and is not related to my changes.
It would actually be good to see whether you are able to reproduce it or whether I did something stupid. So far, I couldn't fully understand why this behavior is observed.

hdsassnick:
Ok, that is a bit weird. I actually tested this feature when implementing it.. Do you have `store_calculated_properties` set to `True` on the `Structure`s? It may also be related to RAM or the number of structures?

I can do some testing next week..

t-reents:
Indeed, I was also confused. Yes, `store_calculated_properties` is `True`, as it's also the default. 

>It may also be related to RAM or the number of structures?
I think I had around 300-400 structures in that example (I'd need to check again)

hdsassnick:
ok, I can actually reproduce the issue. In case of `n_proc=1`, there should be hardly a difference. Since the structure comparison method is taking the pre-calculated property in any case.

For `n_proc>1` the problem should be that an object altered by one processor cannot be shared with the others. That is why pre-calculation should be quicker as all processes access the pre-calculated properties. Maybe, getting a deepcopy of extras is taking more time than recalculating it.. Also, due to the formula check from #61 less structures are actually compared.

In any case, if pre-calculation is not giving an advantage we can omit it in the redesign.

t-reents:
Yes, makes sense

t-reents:
Completed in #78

PR #72, opened 2024-09-02
Unify and simplify interface of `Structure` and `StructureOperations`

t-reents:
As described in #68, I implemented a few changes to simplify the interface between `Structure` and `StructureOperations`.

The PR contains a Jupyter notebook showing that the desired interface actually works.

One thing to discuss: To keep the previous behavior, I had to introduce a different handling of single key/label indices. More precisely, if `append_to_coll == True` and you select a single key, a `StructureOperations` object is returned. However, if `append_to_coll == False`, the same indexing would simply return the single `Structure` object.

On the one hand, this might be considered as slightly inconsistent. On the other hand, it does actually make sense to only return the Structure if you don't want to add it to the initial `StructureCollection`. In case you want to add it, one can argue that the returned `StructureOperations` object is indeed the correct object to reflect it.

Comments are more than welcome.


hdsassnick:
Sorry @t-reents, I am having a quite important deadline coming up on the 11th September. I will try to have a look as soon as possible but not sure if I can make it before that date...

aim2dat/strct/structure_operations.py
hdsassnick, line 1: ``len(self)`` would need to be replaced with ``len(self.structures)``.

hdsassnick (changes requested):
Hi @t-reents, just checked the new implementation and it looks good.

Two things I am not so sure about:

- Would it make sense to implement ``StructureOperations`` as a function? The problem right now is that it kind of is a wrapper for ``StructureOperations``. It is maybe not so straight-forward to do changes or get properties within ``StructureOperations`` on the ``StructureCollection`` object. I guess there are pros and cons for both ways, but maybe we can think about "class vs function" at this point before changing the interface.
- I think we should include this interface change in v0.2.0 (even if it means that this would postpone the new release) to be done with all the breaking changes.

t-reents:
Hi @hdsassnick no worries regarding your deadline, please take your time. Thanks for giving it a look nonetheless!

How would the interface look like when using a function? How do you imagine to call a specific method on the underlying structures etc.?

Something like the following where the method is specified as the name (str)?
```
structure_operations_func(
    strct_coll,
    method_name,
    **dwarfs
)
```


I agree to think about those changes before doing a new release, to avoid breaking changes again and again.

hdsassnick:
> Hi @hdsassnick no worries regarding your deadline, please take your time. Thanks for giving it a look nonetheless!
> 
> How would the interface look like when using a function? How do you imagine to call a specific method on the underlying structures etc.?
> 
> Something like the following where the method is specified as the name (str)?
> 
> ```
> structure_operations_func(
>     strct_coll,
>     method_name,
>     **dwarfs
> )
> ```
> 
> I agree to think about those changes before doing a new release, to avoid breaking changes again and again.

Yes, something like that with individual functions for manipulation, analysis, comparison and identification/removal of duplicate structures. The advantage would be that we are not "wrapping" ``StructureCollection``, so we do not have to care about exposing an interface to it. On the other hand, class properties are not available. Not sure if there are any other downsides to the function approach.

t-reents:
Personally, I find the class approach more intuitive, in the sense that we provide an interface that allows to work with multiple structures in the same way as one would do with a single `Structure`.

One thing that would simplify the logic, at least in my opinion, is removing the logic of appending to the initial `StructureCollection`. In case of analysis methods, we are fine anyway, since they don't produce new structures. However, when manipulating structures, we would simply return a `StructureCollection` containing these results. The user would need to take care of how to combine it with existing data, which should be fine. 

Moreover, when working with a fixed set of structures, I think it feels natural to have this interface that allows to directly perform different operations on it. When using functions, you would always need to pass the same underlying dataset.
Finally, when removing the `append` logic, I think that we don't rely that much on `StructureCollection` which would make the logic quite simple. We only use it as the internal container that is iterated and provide access to the methods of the `Structure` objects.

hdsassnick:
yes, sounds good. Then let us go for the class implementation without ``append`` functionality.

hdsassnick:
Hi @t-reents, just wanted to quickly check on the status of the PR. Would be also happy to help with the implementation in case you are currently busy...

t-reents:
Hi @hdsassnick, thanks for pinging me. Indeed, I was a bit busy with other things, but I picked up the work on this PR this morning. I basically finished the implementation (removing the append logic and small refactoring). I've also fixed the majority of tests. Only a few tests are failing due to my small refactoring. I'll try to fix them this evening.

In case I don't manage to fix those tests, I'll let you know.

Sorry for the delay!

t-reents:
Fixed the tests. You can have another look now, @hdsassnick

hdsassnick:
> Fixed the tests. You can have another look now, @hdsassnick

Hi @t-reents, I believe you forgot to push the commit?

t-reents:
Oh wow 🤣 I also pushed it now

hdsassnick:
Hi @t-reents ,

while trying to fix the ``ml`` sub-package, I encountered an issue:

The output of applying an operation on one structure divers to the application on several structures, as the function is called on the ``Structure`` object itself. I guess this is intended, but it makes the interface more cumbersome as the two cases need to be distinguished. I am wondering whether it would be more intuitive to always return a dictionary/pandas data frame regardless of the length of the key list?

The individual output can still be accessed in any case via: `strct_op.structures[key].function()`

What do you think?

t-reents:
Totally agree. I was also thinking about it while working on this PR. For the historical context, the idea was that when you have a `StructureOperations` object but do the desired operation only on a single structure, the user could expect to get the same output as if it has been done on a "pure" `Structure` object. 

However, one can also argue that a different output from a single object, namely `StructureOperations`, might be even more confusing. Thinking about it now, I would also say that it is more important to have a uniform output for `StructureOperations`, as it simplifies the usage and also development of methods that rely on it. 

I'll fix it in this PR (if you agree?), so basically removing the following parts:

https://github.com/aim2dat/aim2dat/blob/d435beb7e38700c47c2c1d2f2889cd99cad0875e/aim2dat/strct/structure_operations.py#L957-L958

https://github.com/aim2dat/aim2dat/blob/d435beb7e38700c47c2c1d2f2889cd99cad0875e/aim2dat/strct/structure_operations.py#L963-L964

hdsassnick:
That would be great, thank you.

I can then fix the remaining tests and I would also double-check and adapt the documentation.

t-reents:
Fixed it.

Ah true, didn't think about the docs, and also forgot to check the other tests, I was only focussing on the `strct` module. Sorry! If you have the time to fix it, great, but otherwise, please let me know. Just to avoid that we are both working on the same.

hdsassnick (approved):
Hi @t-reents,

as promised, I fixed the tests (the Materials Project tests are probably failing due to the missing keys that are supplied via github secrets) and adjusted the documentation.

If you agree I would merge this PR and it should also close issue #70 and #68, right?

Closes #68

And it should also pave the path towards release v0.2.0. Please let me know if there are any other urgent changes that should be considered....

@Scemp, please let me know if there any objections from your side..

hdsassnick:
Closes #68

t-reents:
Thanks a lot, @hdsassnick!

#70 is not affected by this PR. I only tested the parallelization of the comparison methods in a branch of my fork. If you would like to include those changes in the new release as well, I could try to wrap it up today or tomorrow. In principle, I would only need to extend it to the other comparison methods. I only checked it for the f-fingerprints at the moment.

I'd suggest to do it in a separate PR and merge this one first.

PR #76, opened 2024-09-13
75 import cp2k pdos lists

### LINKED ISSUE #75: Importing pdos lists created with cp2k
At the moment, if you want to import `pdos` for `cp2k`, you only have the opportunity to import it for each element. If you modify the input `LDOS` from cp2k to sum certain atoms, you can not import these lists yet. This is imported in the following line:

https://github.com/aim2dat/aim2dat/blob/eae8d8f4f2045f9b74332295f848a0f366d17dc3/aim2dat/io/cp2k/bands_dos.py#L70

I would suggest changing it to:

```suggestion
@read_multiple(r".*-?(?P<spin>[A-Z]+)?_?(list)(k)?.*\.pdos$")
```
Therefore, the user can also import these modified lists.

Scemp:
closes #75


hdsassnick:
@scemp, thanks, looks good to me.

Ready to be merged?

PR #77, opened 2024-09-16
Bug fix for importing correct data for each plot label in partial charges plot

Scemp:
If a plot label was part of a substring of another plot label, it was added to the plot data. For example, if the functional `PBE` and `PBE0` were plot labels, the data of `PBE0` was included to the `PBE` data. The change differentiates these two as separate plot labels and hence produces the desired plot data.


t-reents:
All good 👍

Issue #16, opened 2024-03-14
Allow list of keys as input of comparison methods in ``StructureOperations``


hdsassnick:
It would be nice to have lists of keys as input for the comparison methods as well. The output could be organized likewise to the one of the distance calculation in the ``Structure`` class.

PR #78, opened 2024-10-10
Parallelize comparison methods of `StructureOperations`

t-reents:
This PR enables the comparison of multiple structures at the same time. Moreover, the `compare_structures_via<method>` methods do also support parallelization now. 

Regarding the comparison of multiple structures:

- If two lists of keys with the same lengths are provided, a pairwise comparison is performed
- If the lengths differ, `itertools.product` is used to perform all possible pairwise comparisons

@hdsassnick I also adjusted the methods in the `ml` module, to take advantage of the new changes. Please let me know if you want to change something.

### DIFF
--- a/aim2dat/ml/cell_grid_search.py
+++ b/aim2dat/ml/cell_grid_search.py
@@ -309,18 +309,18 @@ class CellGridSearch:
 
     def _compare_with_target_structure_ffprint(self, labels):
         scores = []
+        comparisons = self._strct_ops.compare_structures_via_ffingerprint(
+            labels,
+            "target",
+            r_max=self.ffprint_r_max,
+            delta_bin=self.ffprint_delta_bin,
+            sigma=self.ffprint_sigma,
+            use_weights=self.ffprint_use_weights,
+            distinguish_kinds=self.ffprint_distinguish_kinds,
+        )
+
         for label in labels:
-            scores.append(
-                self._strct_ops.compare_structures_via_ffingerprint(
-                    label,
-                    "target",
-                    r_max=self.ffprint_r_max,
-                    delta_bin=self.ffprint_delta_bin,
-                    sigma=self.ffprint_sigma,
-                    use_weights=self.ffprint_use_weights,
-                    distinguish_kinds=self.ffprint_distinguish_kinds,
-                )
-            )
+            scores.append(comparisons[(label, "target")])
         return scores
 
     def _get_model_predictions(self, labels):
--- a/aim2dat/strct/structure_operations.py
+++ b/aim2dat/strct/structure_operations.py
@@ -27,13 +27,24 @@ from aim2dat.strct.strct_comparison import (
 from aim2dat.strct.stability import _calculate_stabilities
 from aim2dat.strct.strct_coordination import _coordination_compare_sites
 from aim2dat.strct.strct_prdf import _ffingerprint_compare_sites
-from aim2dat.strct.strct_space_groups import determine_space_group
-from aim2dat.strct.strct_prdf import calculate_ffingerprint
 import aim2dat.utils.chem_formula as utils_cf
 
 
-def _create_index_combinations(confined, strct_c):
+def _create_index_combinations(confined, strct_c, explicit_indices=None):
     """Create index combinations for duplicate identification."""
+    if explicit_indices is not None:
+        idx1 = explicit_indices[0]
+        idx2 = explicit_indices[1]
+
+        if isinstance(idx1, (int, str)):
+            idx1 = [idx1]
+        if isinstance(idx2, (int, str)):
+            idx2 = [idx2]
+        if len(idx1) != len(idx2):
+            return list(itertools.product(idx1, idx2))
+        else:
+            return list(zip(idx1, idx2))
+
     strct_c_len = len(strct_c)
     if confined is None:
         return list(itertools.combinations(range(strct_c_len), 2))
@@ -303,15 +314,28 @@ class StructureOperations(AnalysisMixin, ManipulationMixin):
         distance : float
             Measure for the similarity of the two structures.
         """
-        return _compare_structures_ffprint(
-            self.structures[key1],
-            self.structures[key2],
-            r_max,
-            delta_bin,
-            sigma,
-            use_weights,
-            use_legacy_smearing,
-            distinguish_kinds,
+        comp_kwargs = {
+            "r_max": r_max,
+            "delta_bin": delta_bin,
+            "sigma": sigma,
+            "use_legacy_smearing": use_legacy_smearing,
+            "distinguish_kinds": distinguish_kinds,
+        }
+
+        index_pairs = _create_index_combinations(None, self.structures, (key1, key2))
+        comp_kwargs["use_weights"] = use_weights
+
+        comp_output = self._compare_structures(
+            compare_function=_compare_structures_ffprint,
+            comp_kwargs=comp_kwargs,
+            threshold=None,
+            index_pairs=index_pairs,
+            desc="ffprint_comp",
+        )
+
+        return self._parse_output(
+            {(idx[0], idx[1]): comp for idx, comp in zip(index_pairs, comp_output)},
+            "ffprint_comp",
         )
 
     def compare_structures_via_comp_sym(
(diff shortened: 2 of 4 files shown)


hdsassnick:
Hi @t-reents,

thank you for the revision as well as addressing issue #16. Everything looks very good to me.

I am just wondering whether it is not too confusing to implement two different behaviours depending on the lengths of the lists. Originally I thought that we could keep the interface the same way as the ``calculate_distance`` methods but I also see the advantage of the "product" approach.

As a third option: we could make the second list optional, if two lists are passed on, pairs from each list are taken, and if just one list is passed on, itertools.product is used. While the first use case gives more flexibility, the second use case would give the output for a similarity matrix.

In any case, I would also strongly opt to keep the ``calculate_distance`` and the ``comparison_*`` methods on par in terms of interface and behaviour. So, if we come to a different conclusion here, I would open an issue to change the other methods...

What do you think?

t-reents:
I definitely agree that the interface should be the same. I just implemented one approach here which seemed to be handy. 

I would prefer your suggestion, i.e. making the second one optional. The motivation would be that the present `itertools.product` approach is very useful for lists of different lengths, but if you want to calculate a similarity matrix, the user will need to use `itertools.product` in advance anyway, as lists of the same lengths are only considered pairwise. Therefore, we would have a different behavior based on the input length.

If the user wants to mimic my present implementation, this can be easily done in place by passing `list1 * 2` and `[idx1] * N + [idx2] * N`.

hdsassnick:
Thanks, sounds good. Please let me know if I can be of any help..

t-reents:
You can have a final look @hdsassnick. I think the methods were not explicitly mentioned in the documentation, so I didn't change anything there. Do you want to explain the logic somewhere? We might want to adjust the doc string mentioning that the second keys/labels input are optional?

Finally, I didn't make the second input explicitly optional (i.e. by providing a default value), as this would require to move it to the last position of the comparison methods arguments. Any opinion on this?

t-reents:
I've already adjusted `calculate_distance`, no?

 I wasn't aware that `calculate_angel` also has the possibility to calculate multiple angles at the same time (although it obviously makes sense to have it there as well)

t-reents:
You are right, the indices generated by itertools were tuples which broke the code. This should be fixed now. 

There is only one remaining point from my side: If the user provides explicit index pairs, it is clear how the output relates to those pairs. However, when the index pairs are generated via itertools, it wouldn't be clear in the current version to which index pair the output belongs. Should we add them as outputs?
Moreover, I switched to `itertools.combinations` as comparison methods and distances are symmetric, and we should avoid doing comparisons with the structures themselves.

Finally, I think it might be already relevant for the current version on main, we don't output the positions in case of `use_supercell = True`, so it's not clear how exactly the outputs are related to the replica? Sure, one knows the pairs of elements involved, but not their positions etc.

t-reents:
> I would add the index pairs as output to make things as consistent as possible. Maybe we could do the following: if single indices/keys are given (without list/tuple), we just return the output value and if lists/tuples are given, the function returns a dictionary with the index/key tuples.

In general, I'd say this makes sense. However, this would basically introduce the same problem that we recently discussed in the refactoring of `StructureOperations` #72, no?. There, we removed the different output formats in case one provides a single key/label. I think the same applies for `calculate_distance` etc., as an interface to `calculate_distance` is also provided via `StructureOperations`.

I think that one can always argue that it is weird to get a dictionary output for something like `structure.calculate_distance(1, 2)`. However, if we simply return the value, we will have the problem again that we have different return formats depending on the input (and we tried to avoid it in the previous PR). 
I'm not saying that I'm against this per se, but if we have this behavior here, I don't see a reason why we should not have it for the comparison methods?
 
>I would transfer the current behavior from calculate_distance for a single index (e.g. just returning the value) to the comparison methods

So does this mean you would revert the change that we did in #72 with respect to the output format?

hdsassnick:
> > I would add the index pairs as output to make things as consistent as possible. Maybe we could do the following: if single indices/keys are given (without list/tuple), we just return the output value and if lists/tuples are given, the function returns a dictionary with the index/key tuples.
> 
> In general, I'd say this makes sense. However, this would basically introduce the same problem that we recently discussed in the refactoring of `StructureOperations` #72, no?. There, we removed the different output formats in case one provides a single key/label. I think the same applies for `calculate_distance` etc., as an interface to `calculate_distance` is also provided via `StructureOperations`.
> 
> I think that one can always argue that it is weird to get a dictionary output for something like `structure.calculate_distance(1, 2)`. However, if we simply return the value, we will have the problem again that we have different return formats depending on the input (and we tried to avoid it in the previous PR). I'm not saying that I'm against this per se, but if we have this behavior here, I don't see a reason why we should not have it for the comparison methods?
> 
> > I would transfer the current behavior from calculate_distance for a single index (e.g. just returning the value) to the comparison methods
> 
> So does this mean you would revert the change that we did in #72 with respect to the output format?

I actually believe that the opposite is true. The suggested changes make it more uniform with regard to #72 as we have the current behavior for ``StructureOperations``:

- ``strct_op[0].determine_space_group()`` returns merely the output, be it a number or in this case a dictionary.
- ``strct_op[[0]].determine_space_group()`` returns the output embedded in a dictionary or a pandas data frame.

I would like to suggest the same behavior for this PR: Two single indices (given as numbers or str in terms of labels) as input will result in the simple output while a list/tuple of indices will result in an output embedded in a dictionary. The problem we encountered in #72 was due to a different behavior for the list/tuple input case with just one element. Here, the output depends on how the input is given but not on its dimension/length. I think this would be in general a suitable and intuitive approach for cases where we allow single values and lists as input.

t-reents:
Well, this was my mistake. As stated before, I also find the "single" return values very intuitive. I just got confused by the changes of the previous PR, I thought we had fully dropped the "different" outputs. Bu you are right, for a single index, we simply return the `Structure` from `StructureOperations`. 

This being said, I agree to unify the interface as you suggested.

t-reents:
Hi @hdsassnick, I actually worked on the PR already now. I added the changes concerning the outputs as discussed. The corresponding tests do also pass.

There is still one problem: The tests concerning the calculation of angles are failing because I unified the output parsing. Therefore, the private methods return the initial lists, even in case of `int` indices. The conversion is only done in `calculate_distance` now. Why are we using the `_calculate_distance` method to simply return the position here? 

https://github.com/aim2dat/aim2dat/blob/159d83bc6868105b1a7f170c1ebbdcefce720667/aim2dat/strct/strct_misc.py#L49-L50

If we were to use `calculate_distance` there, we would need to also return the position for the unit-cell case. But I don't see a reason why the user would expect it. Any suggestions?

hdsassnick:
> Hi @hdsassnick, I actually worked on the PR already now. I added the changes concerning the outputs as discussed. The corresponding tests do also pass.
> 
> There is still one problem: The tests concerning the calculation of angles are failing because I unified the output parsing. Therefore, the private methods return the initial lists, even in case of `int` indices. The conversion is only done in `calculate_distance` now. Why are we using the `_calculate_distance` method to simply return the position here?
> 
> https://github.com/aim2dat/aim2dat/blob/159d83bc6868105b1a7f170c1ebbdcefce720667/aim2dat/strct/strct_misc.py#L49-L50
> 
> If we were to use `calculate_distance` there, we would need to also return the position for the unit-cell case. But I don't see a reason why the user would expect it. Any suggestions?

Because we would like to have the position of the closest replica in case the two sites are separated by a periodic boundary.

t-reents:
Alright, this was also my expectation. I'm not 100% whether this might be better suited in a separate function, but I'd keep it for now and only add the feature that the position can also be returned for the non-supercell case

hdsassnick (approved):
@t-reents, I just had a look at the implementation.

I noticed that the ``compare_*`` functions would still return a dictionary for single indices. So, I changed it to single values. At the same time I tried to make things a bit more slim by moving a few lines of code into the ``_compare_structures`` function... I hope this is ok for you.

Please have a careful look and let me know if you agree with the changes, from my side this PR would be ready to be merged.

t-reents:
>At the same time I tried to make things a bit more slim by moving a few lines of code into the _compare_structures function... I hope this is ok for you.

Sure, I was thinking about the same in the beginning, not sure why I haven't done it in the end. Probably it's just a relic, as I initially used these changes as a proof of concept and didn't clean it up properly.

Thanks for taking care of this. Feel free to merge if nothing else is missing.

t-reents:
> from my side this PR would be ready to be merged.

Just saw your comment. I'll merge it then

PR #80, opened 2024-10-20
Move element data into python modules.

hdsassnick:
In the previous implementation each time the element property is demanded, the corresponding yaml-file is read which can accumulate to a large run time. This PR moves all element data into python dictionaries and thus prevents this issue.


t-reents (approved):
No comments, good to go.

I was wondering, have you seen this approach often before? Just out of curiosity. I was just thinking if it slows down imports or might cause conflicts as "global" variables, but  since we only import it in one place, I think everything is fine.

hdsassnick:
ASE is storing element properties in a similar way. I think the data is still small enough to not affect import times in general.

Of course, for larger data sets we should be more careful..

PR #84, opened 2024-10-25
Update strct_manipulation.py:

jsantana302:
- Add apply_pressure_and_strain function.
- Modify scale_unit_cell function to support scaling_factors as an array.
- Include example notebook for usage.


hdsassnick:
Hi @jsantana302 ,

thank you for the contribution. On first sight, there are a few points that I would suggest to be addressed or think about:

* It would be good to also adjust the tests to cover the new features, e.g. the ones in ``tests/strct/test_structure_op_structure_manipulation.py``
* I am wondering whether we would need a new manipulation function or if we could instead just add more input parameters to the `scale_unit_cell` method to enhance its functionality?
* If you need a new function, you either have to add it to the `ManipulationMixin` class in `strct/mixin.py` to be directly used as a function of the ``Structure`` class (as, e.g., ``scale_unit_cell``) or it should be moved into `strct/ext_manipulation`.
* The example notebook is already a good start but in the final version it should be adopted to match the other notebooks and needs a description (this can be done once the first three points are finalized, I just put it here to not forget about it...).

jsantana302:
Enhanced scale_unit_cell with additional input parameters to support new functionality, avoiding the need for a separate function. Updated tests in tests/strct/test_structure_op_structure_manipulation.py to cover new features and verified all pass with pytest. Revised example notebook to align with existing format.

aim2dat/strct/strct_manipulation.py
```diff
@@ -120,9 +120,52 @@ def substitute_elements(
...
+    from aim2dat.strct import Structure
+
+    def get_strain_tensor(strain):
+        """Convert strain input into a 3x3 strain tensor."""
+        if isinstance(strain, (float, int)):
+            return np.eye(3) * (1 + strain)
+        elif isinstance(strain, list) and len(strain) == 3:
+            return np.diag([1 + s for s in strain])
+        elif isinstance(strain, np.ndarray) and strain.shape == (3, 3):
+            return np.eye(3) + strain
```
t-reents, line 141: I think it's a bit confusing to distinguish between `list`s and `np.array`s like this. If we allow both of them `strain: Union[float, np.ndarray, list]`, I'd suggest to change the `get_strain_tensor` logic. Anisotropic Strain can be provided as a `list` or 1D-`np.array` of length 3. Similarly, a `3x3` strain tensor can be provided as a nested `list` or 2D-`np.array`.

jsantana302: Thanks a lot for your feedback! I modified the function to make it simpler and more concise.

aim2dat/strct/mixin.py
```diff
@@ -449,23 +449,56 @@ def delete_atoms(
     @manipulates_structure
     def scale_unit_cell(
         self,
-        scaling_factor: float = 1.0,
-        change_label: bool = False,
+        scaling_factors: Union[float, List[float]] = None,
+        pressure: float = None,
+        bulk_modulus: float = None,
+        change_label: bool = True,
     ) -> Union["Structure", "StructureCollection"]:
```
Scemp, line 456:
```suggestion
    ) -> Union[Structure, StructureCollection]:
```

This should be type: `Structure` or `StructureCollection` if you use `""` this is considered as a `str` in the list.

Scemp (dismissed):
Apart from the minor adjustment, you are good to go, at least from my side! 

Also, you need to update the `CHANGELOG`.

tests/strct/test_structure_manipulation_functions.py
```diff
@@ -95,3 +97,59 @@ def test_add_structure_random_molecules_error():
...
         == "Could not add guest structure, host structure seems to be too aggregated."
     )
+
+
+def test_scale_unit_cell_uniform_scaling():
+    """Test scale_unit_cell with uniform scaling factors."""
+    structure = Structure.from_file(STRUCTURES_PATH + "MOF-5_prim.xsf")
+    scaling_factors = 1.1
+    scaled_structure = scale_unit_cell(structure, scaling_factors=scaling_factors)
+    expected_cell = np.array(structure["cell"]) * scaling_factors
```
Scemp, line 107: For the `expected_cell` I am not sure if we should have a reference structure or if the function is enough. Maybe @hdsassnick can clarify this. In general, I am not a big fan of testing a function with the same function, but in this case it could be fine. 

This thought is for all `expected_cell`

t-reents: Even if I'm not @hdsassnick, I still put in my 2 cents 😄 

I think the approach is perfectly fine. In this way, you actually test that the function is doing what you expect, since you manually implement the required step again.

Scemp: @t-reents sorry, I didn't want you to feel left out :D 

Yeah, you are correct. Also, if the structure changes (I do not know why we should change `"MOF-5_prim.xsf"` but you never know) the function would still work with another function. So, yes, it should be fine and you can ignore my comment!

aim2dat/strct/mixin.py
t-reents: Exactly, everything is fine from my side. Moreover, I only added the links in case anyone is curious. 

I just created an issue to make things consistent.

hdsassnick (changes requested):
Thank you @jsantana302, looks very good now.

There are some minor things that I would suggest to change. In principle, we should also test for the errors that are raised. If you like you can write an issue and leave this to me.

Also, the example notebook still needs some adjustments and needs to be embedded into the documentation. I will open an issue for that as well and take care of it at a later point.

aim2dat/strct/strct_manipulation.py
```diff
@@ -114,15 +111,58 @@ def substitute_elements(
...
+                isinstance(val, (float, int)) for val in scaling_factors
+            ):
+                return [[scaling_factors[i] if i == j else 0 for j in range(3)] for i in range(3)]
+            if len(scaling_factors) == 3 and all(
+                isinstance(row, list) and len(row) == 3 for row in scaling_factors
+            ):
+                return scaling_factors
+        raise ValueError(
+            "Scaling factors must be a float, a list of 3 values, or a 3x3 nested list."
+        )
```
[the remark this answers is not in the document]
t-reents: Yes, this is what I had in mind, i.e., unifying the behavior for lists and arrays, while still supporting both of them. Thanks @hdsassnick. Just one additional comment. Your current suggestion would fail in case the scaling factors are a row or column vector, i.e. having a shape of `(3, 1)` or `(1, 3)`. Hence, I'd suggest to change line `141` of your suggestion to `elif scaling_factors.size == 3`. `size` simply counts the number of elements.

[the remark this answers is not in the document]
hdsassnick: Thanks @t-reents for the input. I am just wondering whether there is an actual difference between row and column vectors? The following code snippet outputs True..
```
import numpy as np

a = np.array([1, 2, 3])
a.shape == a.T.shape
```

In any case, I adjusted the suggestion above with the suggested size attribute.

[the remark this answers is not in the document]
t-reents: Hi, indeed, in your test case there is no difference, since `a` is a flat array, i.e. `a.shape --> (3,)`. However, if you have a row vector (which is in principle a 2D array, but I still think that you might want to support it), e.g., `a = np.array([[1, 2, 3]])`, you have a shape of `a.shape --> (1, 3)` and `a.T.shape --> (3, 1)`. Hence, the comparison of the shapes would return `False`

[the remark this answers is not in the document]
hdsassnick: Thanks for the clarification. I just double-checked and the current (edited) version should support the 2D row/column array case as well.

jsantana302:
Thanks a lot for your contribution! I think I've addressed the requested changes in the latest commit. Please let me know if further adjustments are needed ⭐

hdsassnick:
@jsantana302, thanks. There are still some tests that are failing related to your implementation (you can ignore the ``StructureImporter`` tests, I will fix those later). Could you please check and fix them?

Once this is done, I can merge the PR.

jsantana302:
Thanks for the reminder! I’ve fixed those issues and I think that after addressing the `StructureImporter` test, it should be ready to go :)

Thank you all for your valuable input @hdsassnick, @t-reents, and @Scemp!

PR #89, opened 2024-11-04
Add hatch to bar plot

Scemp:
Adds the option ```hatch``` to  ```import_bar_data_set``` in the ```SimploPlot``` class to differentiate bar plots if they share key properties but need to be differentiated somehow.

### DIFF
--- a/aim2dat/plots/base_plot.py
+++ b/aim2dat/plots/base_plot.py
@@ -112,12 +112,12 @@ class _BasePlot(abc.ABC):
         "_add_hline",
         "_add_text",
     ]
-    _default_colors = [f"C{idx}" for idx in range(10)]
+    _legendolors = [f"C{idx}" for idx in range(10)]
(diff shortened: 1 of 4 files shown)


hdsassnick:
Nice, thank you for the addition. We would need to add it here as well: https://github.com/aim2dat/aim2dat/blob/main/doc/source/plots-properties_and_functions.ipynb

Scemp:
@hdsassnick `custum_hatch` is added to `doc/source/plots-properties_and_functions` now. I also added `show_colorbar` since it was missing. We discussed adding also `alpha` and `hatch` itself, but it is not a property in general of the `base_plot` class. It is an argument when you import it, for example with the function `import_scatter_data_set`. Thus, I did not add it.

hdsassnick:
Thanks @Scemp, from my side this is ready to go. @t-reents any thoughts?

t-reents:
Nothing to add from my side, thanks!

PR #90, opened 2024-11-05
Fix ``calculate_distance`` function.

### LINKED ISSUE #87: ``calculate_distance`` throws in some cases an error if ``user_supercell=True``
In case the distance between the two atomic sites is larger than ``r_max`` the function throws an error due to recent changes implement by myself.

In this case the output should be ``None`` instead.

hdsassnick:
Closes #87

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -35,7 +35,7 @@ Version 0.2.0
 
 **Depreciations:**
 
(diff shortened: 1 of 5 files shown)


t-reents (approved):
LGTM!

Scemp (approved):
LGTM!

PR #92, opened 2024-11-15
Speed-up n-nearest-neighbours method to calculate coordination and user interface updates.

hdsassnick:
Using ``np.argsort`` is a lot quicker than zipping and sorting for large structures. In addition the user interface of the functions/classes relying on ``calculate_coordination`` are updated.


t-reents (approved):

Scemp (approved):
LGTM

PR #93, opened 2024-11-26
Implement new base block parser and revise cp2k output parser.

hdsassnick:
The original implementation of the CP2K output parser relied on a very complex code structure. The last revision changed the parser to heavily rely on regex patterns, simplifying its code structure.

This revision switches back to normal str recognition and line-by-line parsing but keeps the simple and modular structure of different "blocks" that can be freely combined. Thus, the parser is a lot quicker and the CP2K output parser tests run about 3-4 times faster.

### DIFF
--- a/aim2dat/io/base_parser.py
+++ b/aim2dat/io/base_parser.py
@@ -5,23 +5,56 @@ Parent class for pattern processing and parse function for regex-bases parsers.
 # Standard library imports
 import re
 import abc
-from typing import List
+from typing import List, Union
 
 # Internal library imports
 from aim2dat.io.utils import custom_open
-from aim2dat.utils.dict_tools import dict_set_parameter
+from aim2dat.utils.dict_tools import dict_set_parameter, dict_merge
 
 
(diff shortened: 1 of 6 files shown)


hdsassnick:
The failing tests of the OQMD interface are probably unrelated. Interestingly, the tests run fine on my local computer...

t-reents:
Looks good. Lately, I was also working on parsing for the `aiida-quantuemespresso` plugin. This is something that I also had in mind, but I didn't implement in the end. I might need a tutorial at a some point on how to use it in practice 😄

t-reents (approved):

hdsassnick:
Thanks, in principle, we just have a class for each property/block to be parsed with start and end string. The start string activated the ``parse_line`` function which takes the line and parses the content. Finally, the ``process_output`` function merges all the parsed information into the final output.

After all, this method seems to work very well for typical output files. That being said, it would be great if the DFT codes could support machine-readable output formats such as json, xml, yaml, etc.

t-reents:
Definitely! This something we are also struggling with. Many things can be parsed from the `XML` in `QE`, but certain things are only included in the output file..

Issue #88, opened 2024-11-04
Interface to MOFX-DB


hdsassnick:
Would be nice to have an interface to MOFX-DB in the ``StructureImporter`` class: https://mof.tech.northwestern.edu/

We could either add an interface to search the downloaded database (https://mof.tech.northwestern.edu/databases) or use the REST API or the mofdb_client Python package (https://mof.tech.northwestern.edu/api#versioning).

hdsassnick:
Closed via #94.

PR #94, opened 2024-11-28
MOFX-DB importer

Scemp:
This pull request introduces the functionality to import data from the MOFX database and resolves issue #88.

Currently, this is the initial version, and it is operational. It retrieves structures from the database and generates a new `Structure` for each one downloaded. All the structures are gathered in the `StructureCollection`. Below is an example script:

```(python)
from aim2dat.strct import StructureImporter
strct_imp = StructureImporter()
mofs = strct_imp.import_from_mofxdb(
    vf_min=0.5,
    vf_max=0.99,
    loading_unit="cm3(STP)/cm3",
    pressure_unit="atm",
    database="CoREMOF 2019",
    query_limit=20,
)
```

Given that the structures are imported using `Structure.from_file` and certain space groups may trigger warnings. I'm uncertain about the best approach to take. Should we allow the warning to remain visible to the user, or would it be better to suppress it? Importing 1000 structures could lead to an excessive amount of notifications. However, since this is a warning, it would be beneficial that the user will be informed.

```
UserWarning: crystal system 'triclinic' is not interpreted for space group Spacegroup(1, setting=1). This may result in wrong setting!
  warnings.warn(
```

I must say, I'm not entirely satisfied with how the `heats`, `isotherms`, and `adsorbates` have been integrated into the `attributes` property up to this point. I really need to brainstorm a solution. Perhaps someone out there has a brilliant suggestion?

Looking forward to discussing this pull draft!


aim2dat/ext_interfaces/mofxdb.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def _parse_entry(entry) -> Structure:
+    """Parse entry to structure list."""
+    cif_entry = StringIO(entry.cif)
+    structure = Structure.from_file(cif_entry, file_format="cif")
+    structure.label = str(entry.name)
+    structure._attributes = {
+        "source_id": entry.id,
+        "source": "mofdb-" + str(entry.id),
```
t-reents, line 65:
```suggestion
        "source": f"mofdb-{entry.id}",
```

aim2dat/strct/structure_importer.py
```diff
@@ -226,6 +226,97 @@ def import_from_mp(
...
+        sa_m2g_min: float = None,
+        sa_m2g_max: float = None,
+        sa_m2cm3_min: float = None,
+        sa_m2cm3_max: float = None,
+        name: str = None,
+        database: str = None,
+        telemetry: bool = True,
+        pressure_unit: str = None,
+        loading_unit: str = None,
+        query_limit: int = 1000,
```
t-reents, line 248: I haven't used this database in practice, so I can only comment on the implementation and the expected user experience. Many of the arguments seem to be related to ranges. Have you thought about exposing only the ranges? If a user wants to understand the function and opens the documentation or uses the `?` operator, the amount of inputs might be quite overwhelming. For example: `pld_min` and `pld_max` could only be exposed as `pld` or `pld_range`, i.e. as tuples.

t-reents:
Ah, I forgot your comments:

>However, since this is a warning, it would be beneficial that the user will be informed.

I'd keep the warnings as well. When you use `pymatgen` or any other package, you might also get a bunch of warnings. Typically one can ignore them (or simply does it haha), but it's still good to be reminded that one might want to check certain aspects again. Especially in cases related to symmetry, this might be useful. You receive similar warnings in `seekpath`.

>I must say, I'm not entirely satisfied with how the heats, isotherms, and adsorbates have been integrated into the attributes property up to this point. I really need to brainstorm a solution. Perhaps someone out there has a brilliant suggestion?

I have no spontaneous comment on that, mainly because I don't know how the data would look like. If you have an example to show the point you are not satisfied with, I'm happy to have a look.

hdsassnick:
Hi @Scemp, thanks for the implementation. It already looks very good. I agree with Timo on the ranges, something like ``(min, max)`` would make sense. We can still allow ``None`` in case a limit should not be set.

As for the cif files, you can also use our internal parser by adding ``backend="internal`` to the ``from_file`` function. Would be good to see if there any differences between the two parsers in terms of atom positions and unit cells.

aim2dat/strct/structure_importer.py
Scemp: Thanks for the suggestion, I applied it and you are 100% correct. This implementation made the code much more readable.

Scemp:
@t-reents thanks for the nice suggestions. I implemented them and the code looks much cleaner now. 

> I have no spontaneous comment on that, mainly because I don't know how the data would look like. If you have an example to show the point you are not satisfied with, I'm happy to have a look.

I also brainstormed to retrieve some sort of structure for the data and shifted the metadata to the ``extra`` property. Additionally, ``attributes`` contain the calculated data in form of ``adsorbate`` for ``heats`` and ``isotherms`` including ``pressure`` and ``adsorption`` as lists. 

>I'd keep the warnings as well. When you use pymatgen or any other package, you might also get a bunch of warnings. Typically one can ignore them (or simply does it haha), but it's still good to be reminded that one might want to check certain aspects again. Especially in cases related to symmetry, this might be useful. You receive similar warnings in seekpath.

That is a good point and will leave it like it is!

>As for the cif files, you can also use our internal parser by adding backend="internal to the from_file function. Would be good to see if there any differences between the two parsers in terms of atom positions and unit cells.

@hdsassnick I like your idea to use our internal read function. Since the data is string type and not a file, we need a parser for this. Therefore, I opened a new issue #95 to address this problem.

hdsassnick:
I revised the internal structure and user interface of the new function. At that point I also added tests, updated the CHANGELOG and fixed the tests for the MP interface. From my side the PR could now be merged.

The only failing tests is the one for the optimade interface, I hope this is just temporary...

@Scemp, could you please have look and let me know if you agree with the changes..
@t-reents, if you have some time, please feel free to have a look. Feedback is very welcome.

Scemp:
@hdsassnick thanks a lot for your contribution. I totally agree with your changes, but I made some small adjustments. 

From my side, this PR is good to go!

PR #99, opened 2024-12-22
Fix interface of ``Structure`` to internal cif parser.

### LINKED ISSUE #95: Read structure from string
In the MOFX-DB the CIF is saved as a string. I would suggest adding a method to the function ``get_structure_from_file`` with ``file_format = string``. Otherwise, we need to format the string into a file first to read it.

hdsassnick:
The internal io structure parsers were restriced to be named ``read_*_structure``. This convention clashed with the cif parser being name ``read_file``.

Additionally, support for string as input to ``file_path`` is now given.

Closes #95


Scemp (approved):
Looks good to me! Well done @hdsassnick

PR #101, opened 2025-01-14
Fix optimade interface and update MP test files.

hdsassnick:
This PR merely fixes the currently failing ``StructureImporter`` tests by updating the reference values for Materials Project and making the Optimade interface more stable towards faulty database links.


Scemp (approved):
LGTM!

PR #102, opened 2025-01-16
Add xml parser for Quantum ESPRESSO

hdsassnick:
This PR adds a simple but very complete parser for qe xml output files.


hdsassnick:
Thanks for the approval, I corrected a minor mistake in the structure extraction. If there are no objections, I would then go ahead and merge this one..

PR #104, opened 2025-01-24
Fix io.base_parser.transform_str_value function for loat values and make cif parser more stable.

hdsassnick:
This PR fixes io.base_parser.transform_str_value function for the case that str values would start with a number and were falsely transformed to float and makes the checks for duplicate sites in cif files more thorough.


Scemp (approved):
LGTM!

hdsassnick:
@Scemp, thanks for the review..

I actually just stumbled over the case where multiple sites where duplicate. Now the list of indices is reduced to a set to prevent that the same index is deleted twice.

Scemp:
@hdsassnick you are welcome!

Unfortunately, this did not cross my eyes. Luckily, you saw this and this is a great fix.

PR #105, opened 2025-01-24
Add structure at given position, Rotates sub structure around point or vector

Scemp:
This PR give the opportunity:

- to add a structure at a certain position in the host structure 
- to rotate a certain sub structure, e.g., a molecule integrated in a framework
- to rotate a sub structure around a given vector, e.g., linker molecule in a metal organic framework

### DIFF
--- a/aim2dat/strct/ext_manipulation/add_structure.py
+++ b/aim2dat/strct/ext_manipulation/add_structure.py
@@ -318,7 +318,6 @@ def add_structure_at_position(
     position: List[float],
     guest_structure: Union[Structure, str] = "CH3",
     wrap: bool = False,
-    dist_threshold: Union[float, None] = 0.8,
     change_label: bool = False,
 ):
     """
@@ -337,9 +336,6 @@ def add_structure_at_position(
         or the element symbol to add one single atom.
     wrap : bool (optional)
         Wrap atomic positions back into the unit cell.
-    dist_threshold : float or None (optional)
-        Check the distances between all site pairs of the host and guest structure to ensure that
-        none of the added atoms collide.
 
     Returns
     -------
--- a/aim2dat/strct/ext_manipulation/add_structure.py
+++ b/aim2dat/strct/ext_manipulation/add_structure.py
@@ -312,6 +312,54 @@ def add_structure_coord(
     return new_structure, "_added-" + guest_strct_label
 
 
+@external_manipulation_method
+def add_structure_at_position(
+    structure: Structure,
+    position: List[float],
+    guest_structure: Union[Structure, str] = "CH3",
+    wrap: bool = False,
+    dist_threshold: Union[float, None] = 0.8,
+    change_label: bool = False,
+):
+    """
+    Add structure at a defined position.
+
+    Parameters
+    ----------
+    structure : aim2dat.strct.Structure
+        Structure to which the guest structure is added.
+    position : list of floats
+        Position of the guest structure.
+    guest_structure : str or aim2dat.strct.Structure (optional)
+        A representation of the guest structure given as a string of a functional group or molecule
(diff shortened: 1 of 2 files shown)


aim2dat/strct/ext_manipulation/add_structure.py
```diff
@@ -312,6 +312,52 @@ def add_structure_coord(
     return new_structure, "_added-" + guest_strct_label
 
 
+@external_manipulation_method
+def add_structure_at_position(
```
hdsassnick, line 316: Just a stylistic suggestion, maybe we can choose a shorter name here, eg. ``add_structure_pos`` or ``add_structure_position`` to be in line with the other functions of the module? What do you think?

aim2dat/strct/ext_manipulation/rotate_structure.py
```diff
@@ -0,0 +1,127 @@
...
+
+# Internal library imports
+from aim2dat.strct.ext_manipulation.decorator import (
+    external_manipulation_method,
+)
+from aim2dat.strct.strct import Structure
+
+
+@external_manipulation_method
+def rotate_structure_around_point(
```
hdsassnick, line 21: You think it would be reasonable to merge the two functions into one, maybe something like ``rotate_structure``?

In principle the ``rotation_center`` is the ``origin`` parameter in the second function?

aim2dat/strct/ext_manipulation/rotate_structure.py
```diff
@@ -0,0 +1,127 @@
...
+
+    new_structure = copy.deepcopy(structure)
+    all_positions = list(new_structure["positions"])
+    for idx, pos in zip(site_indices, rotated_points):
+        all_positions[idx] = pos
+    new_structure.set_positions(all_positions)
+    if wrap:
+        new_structure = Structure(**new_structure, wrap=wrap)
+
+    return new_structure, "_rotated-" + f"{angle}"
```
hdsassnick, line 127: If we use the ``to_dict`` function above, we have more control on the transferred properties and we can merge a few lines (same for the other function):
```suggestion
    return Structure(**new_structure, wrap=wrap), "_rotated-" + f"{angle}"
```

hdsassnick (changes requested):
Thanks a lot for the implementation, already looks very good. I fixed the style checks and added a few minor comments.

I am also wondering whether we can use the rotation function for ``add_structure_coord`` to not duplicate code...

Scemp:
Hi @hdsassnick,
thanks for the suggestions. I implemented them and also adjusted the first function `add_structure_random` with the new implemented `rotate_structure` function. While doing, I recognized a bug in the function. The rotation vector was not normalized and thus gave unreasonable results. Since they were constructed random in any case, this should not make a huge difference, but it is fixed now.

Scemp:
Additionally, I need to implement the `rotate_structure` for the other functions in `add_structure` like `add_structure_coord`. 

In addition, I would suggest changing everything which is included in the methods section. The function `add_structure_coord` is really packed... Thus, my idea is to use a dict. For example:

```
def add_structure_coord(
    structure: Structure,
    wrap: bool = False,
    host_indices: Union[int, List[int]] = 0,
    guest_index: int = 0,
    guest_structure: Union[Structure, str] = "CH3",
    guest_dir: Union[None, List[float]] = None,
    bond_length: float = 1.25,
    method_dict: Union[None, dict] = None,
    dist_constraints=[],
    dist_threshold: Union[float, None] = 0.8,
    change_label: bool = False,
)
```

and the `method_dict` could look like:

```
method_dict = {
    "cn_method": {
        "min_dist_delta": 0.1,
        "n_nearest_neighbours": 5,
    }
}
```

With this dict we could write a function grepping the key values and implement them in the `calculate_coordination` function, where there are used. In general, we do not need the method parameters for the `add_structure_coord` function.

Scemp:
Hi @hdsassnick,

all suggestions are implemented, the tests are running well, and I think it is ready to be merged. 

Please take a look, and I am excited to hear your feedback!

PR #107, opened 2025-01-25
Check duplicate positions in strct.Structure validation.

hdsassnick:
This PR introduces a check in the ``Structure`` validation function to check whether the input positions are duplicates of each other.

### DIFF
--- a/aim2dat/strct/strct_validation.py
+++ b/aim2dat/strct/strct_validation.py
@@ -69,10 +69,12 @@ def _structure_validate_positions(positions, is_cartesian, cell, inv_cell, pbc):
     sites = set()
     for i in range(len(pos2check)):
         for j in range(i):
-            if dist_m[i][j] < 1e-3
-                sites.add((i,j))
+            if dist_m[i][j] < 1e-3:
+                sites.add((i, j))
     if sites:
-        raise ValueError("Sites with the same position: " + ", ".join([f"{i}" for i in sites]))
+        raise ValueError(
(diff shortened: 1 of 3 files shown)


Scemp (changes requested):
It already looks excellent and I like the implementation. I only have a small suggestion.

aim2dat/strct/strct_validation.py
```diff
@@ -57,6 +58,19 @@ def _structure_validate_positions(positions, is_cartesian, cell, inv_cell, pbc):
...
+    else:
+        pos2check = []
+        for pos in positions_scaled:
+            pos2check.append([round(pos[i], 15) % 1 if pbc[i] else pos[i] for i in range(3)])
+
+    dist_m = cdist(pos2check, pos2check)
+    for i in range(len(pos2check)):
+        for j in range(i):
+            if dist_m[i][j] < 1e-3:
+                raise ValueError(f"Sites {i} and {j} have the same position.")
```
Scemp, line 72: In this case, we would add each site pair to a list and print the pairs at the end. The idea is that we will find all sites with the same position and do not need to run the script multiple times to find all possibilities.
```suggestion
    sites = set()
    for i in range(len(pos2check)):
        for j in range(i):
            if dist_m[i][j] < 1e-3
                sites.add((i,j))
    if sites:
        raise ValueError("Sites with the same position: " + ", ".join([f"{i}" for i in sites]))
```

Scemp (changes requested):
It already looks excellent and I like the implementation. I only have a small suggestion.

Scemp (approved):
LGTM!

PR #109, opened 2025-01-29
Modified strct-manipulation notebook for better description and visualization

jsantana302:


hdsassnick:
Hi @jsantana302 ,

thank you for the contribution. I have kept all changes except for the method to calculate the coordination environment, "atomic_radius" is quicker and seems to give the same results.

There is still the question whether we can use nglview since the notebooks are also parsed into the documentation: https://aim2dat.github.io/examples.html I will check that at a later point...

jsantana302:
Thanks for your feedback, @hdsassnick. I agree on using "atomic_radius".

I see your point about nglview. However, it could be very useful for interactive visualization. If it doesn’t work properly in the documentation, we could save a static image or embed another viewer as an alternative. Let me know what you find when you check it later. I’d be happy to adjust things if needed.

PR #122, opened 2025-02-21
More units and constants.

### LINKED ISSUE #97: Add more units to units
To extend the analysis for adsorption, pressure, and strain, we have to add additional units in `aim2dat.utils.units`  such as:

- pressure e.g., `"atm"`, `"bar"` (which comes with #94)
- temperature  e.g., `"K"`, `"C"` (which comes with #94)
- adsorbate loading e.g., `"cm3(STP)/cm3"`, `"mmol/g"` (which comes with #94)
- forces e.g., `"hartree/bohr"`, `"eV/angstrom"`

There might be more useful units, but these are the ones I would prefer for now.

hdsassnick:
This PR adds more units and fundamental constants to the ``utils.units`` module based on an internal implementation. Closes #97 

@Scemp , I haven't implemented the units for loadings yet since e.g. ``"cm3(STP)/cm3"`` is not really clearly defined.. maybe this can be done later on..


aim2dat/utils/units.py
```diff
@@ -134,9 +274,22 @@ class Wavevector(_BaseQuantity):
...
+        }
+
 
+constants = Constants()
 length = Length()
 energy = Energy()
+force = Force()
+pressure = Pressure()
 frequency = Frequency()
 wavevector = Wavevector()
```
Scemp, line 294: Maybe is something really simple, but I do not understand why we assign the classes to variables here, and what is the use of this? As far as I can see, we do not use them in this script.

Scemp:
I really like the implementation and that we do not rely on `ase` anymore. Updating the constants is a great choice, and adding the additional parameters will come in handy. 

Regarding the loading units, I am also not sure how to implement them yet. I also have not done any calculations and thus have no feeling about these units. As you suggested, I would leave them for now, and we can implement them later when we know what fits and also which are used by the community.

aim2dat/utils/units.py
hdsassnick: We can do this:

```
from aim2dat.utils.units import energy

print(energy.hartree)
```
Otherwise we would have to do:

```
from aim2dat.utils.units import Energy

energy = Energy()
print(energy.hartree)
```

On the other hand, initiating the classes upon import shouldn't create too much overhead..

Scemp: Thanks for the clarification. Apart from this question, this PR is good to go!

PR #129, opened 2025-03-04
Revision of the external structure manipulation methods.

hdsassnick:
This PR includes:

- The new ``translate_structure`` funciton
- Implements more options for ``dist_threshold`` and implements the parameter in all external manipulation methods.

@Scemp sorry, took a bit longer than anticipated... Could you please check if this resolves your issues?


Scemp (commented):
It looks clean @hdsassnick! 

I only have a few comments, since the Docstrings for the `dist_threshold` parameter does not include `str` as an option, and maybe we need to add `int` as well since it is a valid option. 

Additionally, I added a suggestion to speed up adding the translation vector to the positions, but it is only a minor suggestion.

Overall, well done!

aim2dat/strct/ext_manipulation/translate_structure.py
```diff
@@ -0,0 +1,79 @@
...
+    ValueError
+        If any distance between atoms is outside the threshold.
+    """
+    if site_indices is None:
+        site_indices = list(range(len(structure)))
+
+    vector = np.array(vector)
+    positions = np.array(structure.positions)
+    for idx in site_indices:
+        positions[idx] += vector
```
Scemp, line 74: Damn! I forgot to add it....

```suggestion
    vector = np.array(vector)
    positions = np.array(structure.positions)

    if site_indices is None:
        positions += vector
    else:    
        for idx in site_indices:
            positions[idx] += vector
```

t-reents: I just came along because I saw it in my inbox :D If I understand the logic correctly, you could also use `slice(None)` as the default for `site_indices`. In this way, you don't need any `if-else` logic or loops. 

You should then be able to simply have one line `positions[site_indices] += vector`. If no value for `site_indices` is provided, the default slice is used, and `vector` is added to all rows. If `site_indices` is set to a specific list, only those rows are affected.

hdsassnick:
Thanks for the review, if there are no further comments this branch should be ready to be merged :)

Scemp (approved):
LGTM!

Issue #49, opened 2024-07-08
Add property to `Structure` object to return list of atomic numbers.


hdsassnick:
Sometimes it is handy to have the atomic numbers as a list of integers instead of the element symbols. I would suggest to add a property to the class, e.g., `atomic_numbers` or `numbers` to return the corresponding numbers as tuple.

PR #130, opened 2025-03-05
Add `numbers` property to `Structure` object to return list of atomic numbers.

Scemp:
This PR adds `numbers` as property to the `Structure` object to get atomic numbers. Also adds a dict in `data` where the key is the element with its corresponding atomic number.

Closes: #49


hdsassnick (approved):
Great, thanks @Scemp!

PR #132, opened 2025-03-07
Fix Python 3.13 incompatibility

### LINKED ISSUE #131: Incompatibility with Python 3.13
As class properties are not supported in Python 3.13 anymore, the ``strct.Structure.*_methods`` functions will not work anymore.

Two steps are needed:

1. Fix the source code to be compatible with Python 3.13 and ideally also backwards compatible.
2. Change the Python version in the CI to 3.13.

hdsassnick:
This PR fixes the issues of the source code with Python 3.13. I also implemented minor adjustments to some of the tests and functions.

Closes #131


Scemp (changes requested):
Hi @hdsassnick,

nice implementation and great that you added the tests in the same PR!

I only found some typos. Apart from that, the PR is ready to merch.

Have a great weekend!

hdsassnick:
Thanks @Scemp, I will still wait with the merge until https://github.com/spglib/spglib/issues/544 is fixed as otherwise tests will fail...

Have a great weekend, too!

PR #139, opened 2025-04-15
Fix calculate_distance method.

hdsassnick:
The ``calculate_distance`` method was not back-folding the site position of the "first" indices. This PR fixes it and adds a ``get_position`` function to the ``Structure`` class.


Scemp (approved):
LGTM!

PR #140, opened 2025-04-24
Add error bar to bar plot

Scemp:
This PR introduces error bars to bar plot when importing the data.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,10 @@ Changelog
 Version 0.4.0
 =============
 
+**Enhancements:**
+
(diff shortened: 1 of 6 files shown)


hdsassnick:
@Scemp, thanks for the contribution. On first sight, I am just a bit puzzled why the tests needed to be changed?

Scemp:
Hi @hdsassnick ,
While working on the PR, I noticed that 'y_values' are not needed to import bar data. Thus, I removed the parameter and since it was necessary before, the test failed. I adjusted the test accordingly and removed 'y_values' from the input parameters.

PR #141, opened 2025-05-02
Fix/structure setting cell after initialization and proper handling of pbc of `calculate_distance_function`

hdsassnick:
Just a bugfix for the ``Structure`` class and the ``calculate_distance`` function.


Scemp (approved):
LGTM :+1:

PR #142, opened 2025-05-02
Remove MP legacy interface.

hdsassnick:
This PR removes the legacy interface to MP via pymatgen. Since the implementation in pymatgen has been overhauled an uses the up-to-date openapi interface, the implementation in this library is superseded.


Scemp (approved):
LGTM :+1:

PR #145, opened 2025-05-12
Openmm interface for the ``Structure`` class

hdsassnick:
This PR adds the two functions, ``to_openmm_simulation`` and ``from_openmm_simulation``, to directly create an openmm simulation environment or extract a ``Structure`` object from an openmm simulation.


Scemp (approved):
I am not familiar with openmm, but the implementation looks good! No comments from my side.

PR #150, opened 2025-06-17
added PBE basis sets for Pb and I

hildebell:
Added PBE basis sets for Pb and I since they were missing


hdsassnick:
Thanks for the addition @hildebell

PR #151, opened 2025-07-25
Fix `dist_threshold` in `_check_distances`

Scemp:
This PR addresses the issue where a guest structure cannot be added if the cell is excessively large due to an overly large distance. Previously, the function verified whether **all atoms** in the cell were within the maximum distance between the host and the guest. Now, it ensures that the guest falls within the maximum distance range to the specified indices or to one of the provided elements.

# Example: #
I have a monolayer 12x12 A^2, and I want to add molecules close to the surface. If I set `dist_threshold=[0.8,2.0]` then it was not possible since a few atoms could only be in range of 2 A. Now, if one host atom is in between 0.8 A and 2.0 A and each host atom has a minimum distance of 0.8 A, the molecule is added.


Scemp:
Well, I tried to fix the License error as well but I have no idea on how to. Maybe, @hdsassnick, you have a quick solution :)

hdsassnick (approved):
Ok, I see. The problem with the CI is due to the dscribe package, I will see how I can fix it in a separate PR.

As for this PR, I have adjusted the CHANGELOG, I think this should be considered as breaking change. If you are happy with it I can merge it. Do you think, we would need some tests?

Also, would be nice to have your use-case as example notebook, what do you think?

PR #153, opened 2025-07-30
Enh/adsorption workflow

Scemp:
This PR includes the adsorption of molecules in metal-organic frameworks. In more detail, the following was added:

- Include a WF for adsorption:
  - WF for the molecule (Geo opt)
  - Pre-adsorption for the framework (find equivalent sites, Cell opt, pDOS and partial charges)
  - Adsorption with the option to fix the molecule in framework (Geo opt, pDOS and partial charges)
- New functionals and vdW corrections:
  - R2SCAN-D3 and R2SCAN-rVV10
  - PBE0, PBE0-D3, and PBE0-rVV10 
  - HSE06, HSE06-D3, and HSE06rVV10
- Geo optimization:
  - includes list for fixed atoms
- New calc functions:
  - add_molecule (adds a molecule based on the coordination of the host)
  - find_equivalent_sites (finds equivalent sited of the given structure [used in the adsorption WF])
  - rotate_molecule (rotates the given molecule based on its indices in the framework)


Scemp:
Thanks @hdsassnick! :) Well, when adding the functions, I saw the typos and thought, I might fix them :D

Braking changes are indeed the changes of the basis sets and SCF steps in OT, as well as the change of the bond distance in H₂.

I also adapted the function that if the SCF does not converge a new iteration will start in for the WF `find_scf_parameters`, which is more of a fix.

Apart from that everything is only addition.

I will include everything in the CHANGELOG and will let you know when it is done.

Scemp:
@hdsassnick thanks a lot for the minor adjustment!

I had planned to implement the LDOS instead of using different kinds to calculate the contribution of each atom to the DOS. In the same go, you can use a list to define the contribution of subgroups as well.

Additionally, I adapted the function to add a structure via coordination. 

I reckon I will be done testing by the end of the week, and then it would be great to release the new version. For those contributions, I will open a new PR asap. Also, I know that @hildebell wanted to implement a function that greps the calculation time from the Slurm output if the calculation was done remotely. I will kindly ask her to create a PR :)

hdsassnick:
ok, great. There is no rush. I believe that the calculation time is already parsed from the CP2K output file? It may not be stored in the output-dictionary of the AiiDA calculation, but would be probably the easier way to change that...

PR #157, opened 2025-08-11
Fix CP2k ExitCodes

Scemp:
This PR introduces fixes for the `ExitCode` in CP2K. This includes the following:

- `ExitCode` 610 was changed to 501, since 610 is defined for the `ProcessHandlerReport` and 501 for the `CalcJob`.
- `_resubmit_calculation` when 1. no output was created (this happens when a connection error occurs but is somehow not detected) or 2. when the optimization reaches wall time.
- Add wall time to CP2K input if set for slurm -> Thus, an error can occur for the `_BaseOptimizationWorkChain` class and `_resubmit_calculation` will be performed

### DIFF
--- a/aim2dat/aiida_workflows/cp2k/parsers.py
+++ b/aim2dat/aiida_workflows/cp2k/parsers.py
@@ -84,7 +84,7 @@ class _Cp2kBaseParser(Parser):
             return self.exit_codes.ERROR_ILL_CONDITIONED_MATRIX
         elif "bad_condition_number" in result_dict:
             return self.exit_codes.ERROR_BAD_CONDITION_NUMBER
-        elif result_dict["exceeded_walltime"]:
+        elif result_dict.get("exceeded_walltime", False):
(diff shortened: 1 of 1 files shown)


aim2dat/aiida_workflows/cp2k/parsers.py
```diff
@@ -84,7 +84,7 @@ def parse(self, **kwargs):
             return self.exit_codes.ERROR_ILL_CONDITIONED_MATRIX
         elif "bad_condition_number" in result_dict:
             return self.exit_codes.ERROR_BAD_CONDITION_NUMBER
-        elif result_dict.get("exceeded_walltime", False):
+        elif result_dict["exceeded_walltime"]:
```
hdsassnick: I am not so sure about this change since the original version was safer with regard to changes in the stdout parser. With the change we have to make sure to always have the key `"exceeded_walltime"` in the output dictionary. I leave it up to you whether to keep it or not..

hdsassnick (approved):
Looks good, thanks a lot, just one minor comment.

PR #160, opened 2025-08-12
Fix cp2k core wc handler tests.

hdsassnick:
Fixing some tests that are broken from #157

### DIFF
--- a/aim2dat/aiida_workflows/cp2k/core_work_chain_inputs.py
+++ b/aim2dat/aiida_workflows/cp2k/core_work_chain_inputs.py
@@ -51,6 +51,7 @@ def _set_numerical_p_xc_functional(cp2k_dict, input_p, cp2k_version):
 
 def _set_numerical_p_basis_sets(cp2k_dict, input_p, structure, xc_functional):
     """Validate and set the basis set parameters."""
+    bs_type = ""
     if isinstance(input_p, aiida_orm.Str):
         if xc_functional is None:
             return False, {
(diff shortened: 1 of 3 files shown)


Scemp:
Hi @hdsassnick,

thanks for pointing it out. I totally forgot the test, to be honest. The error occurs because I deleted the check if `scf_converged` is in `output_p_dict` in `_resubmit_calculation` since this should be done here and raises another `ExitCode`: 
https://github.com/aim2dat/aim2dat/blob/4dd3fe406af6d5f625a9dcd6d204c3936bebdfcd/aim2dat/aiida_workflows/cp2k/parsers.py#L77-L78
and will is dealt by another function and is tested here:
https://github.com/aim2dat/aim2dat/blob/4dd3fe406af6d5f625a9dcd6d204c3936bebdfcd/tests/aiida_workflows/test_cp2k_core_wc_handlers.py#L49-L51

Since this is not failing in the new function `_resubmit_calculation`, we cannot get the `ExitStatus 611`.

But in the same go, we should test if the calculation will be resubmitted if the `walltime` is hit. I will change the test later this day.

hdsassnick:
great, thanks. At the moment, priority is that it works in practice. If you can design a test within the current framework this is very welcome but also no problem to just remove the failing test.

I would like to change the AiiDA testing infrastructure at some point in any case.  I would like to design small test cases for every workchain and run them once, export the data into an archive and the caching mechanism can capture the corresponding calcjobs such that no actual calculations need to be performed when running the test CI. Ideally, we would also rely on the new sqlite backend and get rid of PostgreSQL for tests.

Scemp:
@hdsassnick,

the test case where the wall time was hit was added. At the end, it was only a small adjustment, but the test is fixed now.

Regarding your suggestion to reorganize the Aiida test cases, I think that would be an excellent idea! If you require any assistance, feel free to reach out!

hdsassnick:
Perfect, thanks.

PR #168, opened 2025-08-20
Enh/add_structure

Scemp:
This PR enhances the `add_structure_coord` function. This includes:

- if the added `guest_structure` does not satisfy (mostely because of wrong calculated `bond_dir`) a grid search around the `host_atoms` will be performed. The grid is a sphere around the host atoms.
- `constrain_steps` is changed from different numbers to perform the steps to the angles used in this grid search.
- the calculated `score` for this action is moved to `_check_distances` and in case the flag `return_score` is set to `True` a value will be returned instead of `True`.
- based on the previous implementation, a flag `rotate_guest` was added. If it is set to `True` the guest molecule will be rotated around its own axis until the `score` calculated with `_check_distances` is minimal.
- `distance_constrains` was removed because `dist_threshold` covers this case as well.
- `guest_dir` is removed since the algorithm deals with the `guest_structure` the most suitable way possible. If not satisfied, the `guest_structure` can be rotated later with the `rotate_structure` function.
- `guest_index` is now `guest_indices` and multiple indices can be passed now. (In case you want your hydrogen from `H2O` pointing towards the `host_structure` just use `guest_indices=[1,2]`.)
- The `bond_length` is thus also adjusted for this case.
- some test cases had to be changed due to the new implementation of `_derive_bond`. Which now considers all `neighbours` to create `bond_dir`.


hdsassnick (approved):
Thanks @Scemp, looks good.

PR #169, opened 2025-08-25
Add frame to bar plot

Scemp:
Small addition to bar plot.


hdsassnick:
Perfect, thanks @Scemp

PR #171, opened 2025-08-27
Fix/scf conv cp2k

Scemp:
This PR fixes:

- Restart if SCF does not converge.
- Restart if the wall time is hit. 

And extends the SCF iteration.


aim2dat/aiida_workflows/cp2k/calcjobs.py
```diff
@@ -240,9 +240,6 @@ def prepare_for_submission(self, folder):
         """Prepare input for calculation."""
         parameters = _set_dict_to_upper(self.inputs.parameters.get_dict())
         dict_set_parameter(parameters, ["GLOBAL", "PROJECT"], self._PROJECT_NAME)
-        walltime = self.inputs["metadata"]["options"].get("max_wallclock_seconds")
-        if walltime:
-            dict_set_parameter(parameters, ["GLOBAL", "WALLTIME"], walltime)
```
Scemp, line 245: Well, setting the `WALLTIME` in CP2K is not as trivial. The Slurm raises the error before CP2K if the `WALLTIME` is set to the same number. Since the `WALLTIME` is heavily dependent on the functional, the size of the structure, and other parameters, I will leave it out for now unless someone has a better idea.

[https://github.com/cp2k/cp2k/issues/4399](url)

hdsassnick: Ja, I guess it is what it is.. also depends on the used hardware. I would leave it up to the user to set a reasonable walltime or not.

Scemp:
@hdsassnick can you take a look at the test, please?

hdsassnick:
@scemp, looks like an issue with the optimade interface, maybe it is just a temporary problem..

In any case it is unrelated to this PR, so we can merge it if you are happy with it.

PR #173, opened 2025-09-02
Bugfix for ``Structure.to_dict`` function

hdsassnick:
With PR #170 the ``Structure.attributes`` have been made mutable, removing the deepcopy from the getter function. Thus, the ``to_dict`` function delivered mutable output which is heavily used by manipulation functions, leading to undesired behavior.

To fix this the deepcopy has been introduced in the ``to_dict`` function and a test was added.


hdsassnick:
The `StructureImporter` tests are unrelated, I will take care of it if they persist.

Scemp (approved):
LGTM

PR #174, opened 2025-09-05
Fix/chargemol

Scemp:
This PR adds `runtime` to the `chargemol` and `critic2` workchains and also includes fixes for the `chargemol` protocols.

@hdsassnick I was not sure how to handle the other cases since I was not really getting through the compare flag. Maybe you can double-check it, and if possible, provide me with some information about the compare flag.


Scemp:
@hdsassnick 

Thanks for the clarification. In this case, the `'compare'` key does not have to be set. Thus, the PR is ready to be merged if you give the green light.

PR #197, opened 2026-04-15
Remove `get_statistics` kwarg in `calc_coordination`.

hdsassnick:
This PR reverts the implementation of `get_statitistics` from #177. 

Since the function is used for quite a lot of other analysis and manipulation functions, it creates quite a burden to check for the different output options. I also refrained from a deprecation since the kwarg has been introduced very recently and has probably no impact on down-stream scripts/packages.


Scemp (approved):
LGTM

PR #202, opened 2026-05-07
Enh/adsorption

Scemp:
This is still a draft pull, since I would like to add multiple additions:

1. Add charge density difference (CDD) to adsorption workflow to further analyse the electronic properties.
2. If `.Cube` files are already produces (CDD or partial charges) they could be reused for the other calculations.

Additionally, I have some thoughts, but I would like to discuss them:

1. Add BSSE when molecule is added to analyse the partial charge by default or set as optional flag?

This PR addresses the following fixes and enhances:

**Enhancements:**
- New metals are added to ``aiida_workflows.cp2k.parameter_files.basis_sets.yaml``
- New adsorption protocols ``cp2k-adsorption-mofs-standard_v2.0`` and ``cp2k-crystal-mofs_v4.0`` are added to the ``aiida_workflows.protocols.``

**Fixes:**
- The resubmit loop was fixed in ``aiida_workflows.cp2k.base_core_work_chain`` 
- In the BSSE calculation the fragments were fixed ``aiida_workflows.cp2k.interaction_energy_work_chain``, as well as the scf convergence for each calculation in the BSSE WC in ``io.cp2k.stdout``
- The still retrieved runtime error when the last row was empty. Fixed in ``aiida_workflows.utils.get_workchain_runtime``


hdsassnick (approved):
Thanks a lot for the work and contribution. I have just added the new AiiDA processes to the documentation source file. I believe that the test errors are unrelated, I will fix that later...

From my side this one is good to go.

Scemp:
Thanks @hdsassnick for the revision. I also think that the tests are unrelated to this PR.

I will merge it now.

PR #210, opened 2026-07-02
Added finite-displacement phonon workflow for CP2K

Scemp:
This PR is an adaption from #209. A huge shoutout to Maddox for putting this together! 

## Summary

Adds a complete finite-displacement phonon band structure + DOS pipeline
for CP2K, driven by phonopy, and integrated with aim2dat's existing
WorkflowBuilder protocol system. Also rearranged functions for external 
phonon module to work with new implemented workflow.

## Enhancements

`aim2dat.cp2k.phonons` (PhononWorkChain) orchestrates:

1. `phonopy_generate_displacements` → N displaced supercells
2. `ForceWorkChain` × N → CP2K ENERGY_FORCE per displacement
   (inherits _BaseCoreWorkChain SCF auto-recovery)
3. `phonopy_collect_phonons` → band structure + DOS (+ thermal properties)

Outputs: `phonon_bands` (BandsData), `phonon_dos` (XyData),
`phonon_setting_info` (Dict), optional `thermal_properties` (XyData).

## Fixes

- rewrite `ext_interfaces.phonopy._extract_*` to match newest phonopy version

## TO-DO

- adapt the `cp2k-crystal-phonons_v1.0.yaml`
- add `drift_forces` to https://github.com/aim2dat/aim2dat/blob/add-phonon-workflow/aim2dat/aiida_workflows/cp2k/parsers.py#L248-L250 similar to https://github.com/phonopy/phonopy/blob/main/phonopy/interface/vasp.py#L56
- test again
  - Si PBE/TZV2P: full pipeline finished [0] on GWDG, band structure + DOS
  consistent with Materials Project reference (✓)
  - NaCl PBE/TZV2P: full pipeline finished [0], acoustic modes = 0 at Γ, but Imaginary freqs at boundaries and no LO-TO splitting observed - v2 fix (×)

## Further suggestions

- write phonopy as (Calcjob)
  - no need to run locally
  - more flexibility
  - adds transparency (which version was used etc.)
- rewrite `ext_interfaces.phonopy._extract_* since it does not extract the properties; it actually runs the calculation
  - idea would be to add each of these extraction as a possibility to run the Calcjob from above
  - Non-Analytical Corrections(NAC) reserved as `phonopy_p.nac_parameters` input (v2 TODO) —
  API designed so adding LO-TO correction doesn't break existing inputs

## Notes from reviewers

~~1. SeekPath → primitive cell + band path~~ → should be done before the Workchain
~~4. `parse_cp2k_forces` → stacked force sets via phonopy's CP2K interface~~ → is now implemented in the standard CP2K parser


aim2dat/aiida_workflows/cp2k/phonon_work_chain.py
```diff
@@ -0,0 +1,287 @@
...
+1. ``seekpath_structure_analysis``  -> primitive cell + q-point band path
+2. ``phonopy_generate_displacements`` -> N displaced supercells + setting info
+3. ``ForceWorkChain`` x N            -> CP2K ENERGY_FORCE per displacement
+4. ``parse_cp2k_forces``             -> stacked force sets
+5. ``phonopy_collect_phonons``       -> phonon band structure + DOS
+
+Proposed entry point (pyproject.toml)::
+
+    [project.entry-points."aiida.workflows"]
+    "aim2dat.cp2k.phonons" =
```
hdsassnick: For consistency reasons I would suggest "aim2dat.cp2k.combined.phonons" as entry point.

```suggestion
    "aim2dat.cp2k.combined.phonons" =
```

aim2dat/aiida_workflows/cp2k/phonon_work_chain.py
```diff
@@ -0,0 +1,287 @@
...
+
+class PhononWorkChain(WorkChain):
+    """
+    AiiDA work chain to compute the phonon band structure and DOS via the
+    finite-displacement method with CP2K and phonopy.
+    """
+
+    _keep_scf_method_fixed = True
+    _keep_smearing_fixed = True
+    _initial_scf_guess = "RESTART"
```
hdsassnick: This should not be needed since the class does not inherit from ``_BaseCoreWorkChain``

```Suggestion
```

aim2dat/aiida_workflows/protocols/cp2k-crystal-phonons_v1.0.yaml
```diff
@@ -0,0 +1,381 @@
...
+    blacklist_inputs:
+    - custom_scf_method
+    - optimization_p.keep_angles
+    - optimization_p.keep_symmetry
+    dependencies:
+      unit_cell_preopt:
+      - [cp2k.output_structure, structural_p.structure]
+      - [scf_parameters, structural_p.scf_parameters]
+  phonons:
+    process: aim2dat.cp2k.phonons
```
hdsassnick: See comment above.

```suggestion
    process: aim2dat.cp2k.combined.phonons
```

aim2dat/aiida_workflows/utils.py
```diff
@@ -21,11 +23,15 @@
 from aim2dat.strct.analysis.brillouin_zone_2d import _get_kpath
 from aim2dat.utils.dict_tools import dict_retrieve_parameter
```
hdsassnick, line 24: We can use the `_return_ext_interface_modules` function to interface phonopy below:
```Suggestion
from aim2dat.utils.dict_tools import dict_retrieve_parameter
from aim2dat.ext_interfaces import _return_ext_interface_modules
```

aim2dat/aiida_workflows/utils.py
```diff
@@ -163,6 +169,235 @@ def create_surface_slab(surface, nr_layers, parameters):
...
+
+    Returns
+    -------
+    dict
+        ``supercell_0000`` ... ``supercell_NNNN`` : StructureData
+            One displaced supercell per displacement, fed to the per-displacement
+            force calculations (:class:`ForceWorkChain`, RUN_TYPE ENERGY_FORCE).
+    """
+    from aim2dat.ext_interfaces.phonopy import _build_phonopy
+
```
hdsassnick: This would be not needed when using `_return_ext_interface_modules`.

```Suggestion
phonopy_module = _return_ext_interface_modules("phonopy")
```

aim2dat/aiida_workflows/utils.py
```diff
@@ -163,6 +169,235 @@ def create_surface_slab(surface, nr_layers, parameters):
...
+    supercell_matrix = phonopy_dict.pop("supercell_matrix")
+    primitive_matrix = phonopy_dict.pop("primitive_matrix", "auto")
+    symprec = phonopy_dict.pop("symprec", 1e-5)
+    calculator = phonopy_dict.pop("calculator", "cp2k")
+
+    p_dict = parameters.get_dict()
+    displacement = p_dict.pop("displacement", 0.01)
+
+    phonopy_atoms = Structure.from_aiida_structuredata(structure).to_phonopy_atoms()
+    phonon = _build_phonopy(
```
hdsassnick: Using `_return_ext_interface_modules`:
```Suggestion
    phonon = phonopy_module._build_phonopy(
```

aim2dat/ext_interfaces/phonopy.py
```diff
@@ -1,9 +1,17 @@
...
+
 # Third party library imports
 import phonopy
+from phonopy.structure.atoms import PhonopyAtoms
 from phonopy.phonon.band_structure import get_band_qpoints_and_path_connections
 from phonopy.file_IO import read_v_e, read_thermal_properties_yaml
+from aiida.plugins import DataFactory
+
+
+StructureData = DataFactory("core.structure")
```
hdsassnick: Is this needed here? I would like to separate the software packages to be able to use/interface phonopy also without haveing installed aiida.
```Suggestion
```

hdsassnick (dismissed):
Hi @madd-ox and @Scemp,

sorry for taking so long. Thank you, @madd-ox, for such a thought-out implementation. I really appreciate all the work you put into this.

I added a few minor suggestions. We will have to fix the `test_list_methods` test in `tests/strct/test_structure.py`. This can happen within this PR or I can do it later on.

I haven't run the AiiDA workflow and related tests yet since I don't have aiida configured on this computer. Nevertheless, I would be happy to merge it (with the suggested changes) and I could have a look later this month.

aim2dat/aiida_workflows/cp2k/parsers.py
```diff
@@ -241,6 +242,13 @@ def _create_output_nodes(self, result_dict, output_structure):
         if "eigenvalues_info" in result_dict:
             ev_info = result_dict.pop("eigenvalues_info")
             self.out("output_eigenvalues", Dict(dict=ev_info))
+        if "forces" in result_dict:
+            forces = result_dict.pop("forces")
+            forces_array = ArrayData()
+            forces_array.set_array(
+                "forces", np.array([result["atom_forces"] for result in forces])
+            )
+            self.out("output_forces", forces_array)
```
Scemp, line 251:
```suggestion
            forces = result_dict.pop("forces")
            atom_forces = np.array([result["atom_forces"] for result in forces])
            drift = np.mean(atom_forces, axis=0)
            forces_array = ArrayData()
            forces_array.set_array("forces", atom_forces-drift)
            self.out("output_forces", forces_array)
```

aim2dat/ext_interfaces/phonopy.py
Scemp: No, this is actually not needed. In my first attempt, I needed it, but then I moved it and forgot to delete. Thanks for pointing it out.

aim2dat/aiida_workflows/cp2k/parsers.py
madd-ox: I think this would conflict with the check Julia's suggested to make sure phonon input structure has max forces ~0 before generating displacement and running other force calcs (exits with code 402). If we subtract drift on forces automatically before returning I think it would influence this check. So, I am implementing a settings.output_remove_force_drift (default False) so this will not conflict with my existing checks. I also have a subtract_ref_forces param (default False) in the PhononWorkChain which does something similar, but rather than averaging we subtract per-atom forces. I will be committing my current code by end of day

Scemp (commented):
Apply suggestions from code review

PR #213, opened 2026-07-16
Follow-up to phonon workflow: New reference-force gate check and updated protocol to cp2k-crystal-phonons-only_v1.1

madd-ox:
Follow-on to #210 (already merged). Adds reference-force gate check and consolidates the phonon protocol into a single cp2k-phonons_v1.0.yaml:

-  Combined the cell-optimization stage into cp2k-phonons_v1.0 as one protocol (find_scf_parameters -> unit_cell_preopt -> unit_cell_opt -> phonons), instead of chaining it through a separate cp2k-crystal-opt_v1.0 protocol that largely duplicated cp2k-crystal-standard/preopt.
-   Re-architected to SeekPath-first: seekpath_analysis now runs before the cell-opt tasks, so unit_cell_opt relaxes the primitive cell directly rather than optimizing the conventional cell and reducing to the primitive cell afterward. This avoids a later re-expression into a different cell setting, which can change FFT grid registration enough to fail the phonon reference-force gate. 
-   Swapped optimization_p.keep_space_group for optimization_p.keep_angles in the cell-opt tasks: KEEP_SPACE_GROUP's per-step spglib re-detection is fragile on a non-orthogonal cell, while KEEP_ANGLES is a plain geometric constraint with no re-detection step involved. NOT YET VALIDATED: this SeekPath-first + keep_angles combination hasn't been re-run end-to end against Si/NaCl yet -- flagging for review/ discussion before merge, not presenting it as tested.
-   Collapsed find_scf_parameters back to a single task feeding both unit_cell_opt and the phonon fan-out, since the conventional-vs-primitive cell mismatch that justified a second SCF search no longer exists.
-   Reference-force gate and PhononWorkChain docstring/comment cleanup.
-   Bumped the phonons extra's phonopy floor to >=4.2.1 (the version this code is actually written/tested against -- open to discussion, happy to revert to >=2.17.1 if preferred).

 Files touched: phonon_work_chain.py, phonopy.py, parsers.py, utils.py, workflow_builder.py, pyproject.toml; cp2k-crystal-phonons_v1.0.yaml replaced by cp2k-phonons_v1.0.yaml.

Note: Tests on Si and Na-Cl running now. I will add a new comment confirming this new protocol works when these calculations finish


madd-ox:
NaCl and Si are not working with this new pipeline. After doing some trouble shooting and looking into the docs, CP2K does not support cell optimizations on primitive rhombohedral cells of the FCC lattice (such as the case with the seekpath derived primitive cell of NaCl and UiO-66). The pipeline works much better after switching to the conventional cubic cell for cell optimization, running seekpath afterwards to convert to primitive cell/get path params, running a confirmatory geometry optimization on the primitive, and then finally passing the primitive to the phonon work chain.  I will be updating the workflow to match this scheme and pushing a new update soon

madd-ox:
New pipeline uploaded and confirmed working! The new phonon protocol is designed to run post-optimization. This way we can optimize the conventional cubic cell (avoiding CP2K cell optimizer fails for FCC cubic like UiO-66 and NaCl with cp2k-conventional-crystal-opt_v1.0) or utilize whatever other optimization protocol, and then run phonons on the primitive cell for more efficient phonon calculations.

